PR #11, opened 2022-04-22
Dev : added 'dist-upgrade' and 'autoremove' 

ymauray:
In order to alias `apt` to `deb-get`, it needs a few more options.
This pull request adds `dist-upgrade` and `autoremove`.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -4,8 +4,8 @@ LC_ALL=C
 function usage() {
   echo -e "Usage
 
-deb-get {update | upgrade | show pkg | install pkg | reinstall pkg |
-remove pkg | purge pkg | search pkg | cache | clean | list | help | version}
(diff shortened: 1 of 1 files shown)


popey:
Is the intention of deb-get to replace apt? I think it's dangerous to go down the path of intending to alias `apt` to `deb-get`. You'll constantly be chasing things `apt` does that `deb-get` doesn't. For example being able to `apt install foo=1.0` or `apt install foo/pocket`.

flexiondotorg:
It was not my intention to make `deb-get` a complete wrapper/replacement for `apt` or `apt-get` and @popey makes a good point. If `deb-get` operates more like `apt/apt-get` then it sets user expectation that more complete apt/apt-get functionality should work, which is definitely out of scope.

So, I am going to politely decline this pull request. But thank you for contributing :slightly_smiling_face:

PR #12, opened 2022-04-22
deb-get: Add insync

arif-ali:
Add insync

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -717,6 +717,15 @@ function deb_google-earth-pro-stable() {
     WEBSITE="https://www.google.com/earth/versions/"
 }
 
+function deb_insync() {
+    ARCHS_SUPPORTED="amd64"
+    APT_KEY_URL="$(curl -s https://www.insynchq.com/downloads | grep gpgkey | cut -d'=' -f2)"
+    APT_LIST_NAME="insync"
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -113,6 +113,9 @@ function install_apt() {
     if [ -n "${APT_KEY_URL}" ]; then
         wget -q "${APT_KEY_URL}" -O "/etc/apt/trusted.gpg.d/${APT_LIST_NAME}.asc"
     fi
+    if [ -n "${APT_KEY_ID}" ]; then
+        apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "${APT_KEY_ID}"
```
flexiondotorg, line 117: apt-key is deprecated. So an alternative method of acquiring the key is required.

flexiondotorg (changes requested):
`apt-key` is deprecated. The URL to public signing key can be acquired using: `curl -s https://www.insynchq.com/downloads | grep gpgkey | cut -d'=' -f2`

Does this key also authenticate the InSync apt repo? Any existing Insync keys need to be removed from the trusted keyring(s) before this can be tested.

deb-get
arif-ali: okie, didn't realise it was deprecated. Updated PR, with updated URL. tested this in a LXD container, and seems to work for me

PR #14, opened 2022-04-23
Add jabref installer

LyzardKing:
I added the bibtex manager JabRef, which publishes .deb files at https://builds.jabref.org/main/

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -927,8 +927,8 @@ function deb_jabref() {
     get_github_releases "https://api.github.com/repos/jabref/jabref/releases/latest"
     VERSION_PUBLISHED="$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4 | cut -d'/' -f8 | sed 's|v||')"
     URL=$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)
-    PRETTY_NAME="GitHub Desktop"
-    WEBSITE="https://desktop.github.com/"
+    PRETTY_NAME="JabRef"
+    WEBSITE="https://www.jabref.org/"
 }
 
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -922,6 +922,14 @@ function deb_exodus() {
     WEBSITE="https://exodus.com/"
 }
 
+function deb_jabref() {
+    ARCHS_SUPPORTED="amd64"
+    get_github_releases "https://api.github.com/repos/jabref/jabref/releases/latest"
+    VERSION_PUBLISHED="$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4 | cut -d'/' -f8 | sed 's|v||')"
+    URL=$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)
+    PRETTY_NAME="GitHub Desktop"
```
flexiondotorg, line 930: Please correct the PRETTY_NAME and WEBSITE

flexiondotorg (changes requested):
Please date this to correctly reference the JabRef

deb-get
LyzardKing: Right!
I started changing it with the github data, and removed the wrong lines. It should be fixed now

wujastyk:
I would very much like to be able use deb-get to fetch the latest build of JabRef from `builds.jabref.org/main` .  As far as I can see, this is what this installer did, originally, but in April last year it was switched to fetch the public stable release from Github.  Could and installer with the original behaviour be added, perhaps with a different name (`jabref-test` perhaps?).

LyzardKing:
The change was made to make the stable version available from deb-get.
That is the same version that a user would download from the upstream website (https://www.jabref.org/#download).
This is the better default, keeping in mind that it can be overrided (as @philclifford mentioned).

PR #39, opened 2022-05-05
add azure-cli

DavidGoodwin:
add support for the azure-cli tool.


flexiondotorg (changes requested):
Please use `${UBUNTU_CODENAME}` instead of `lsb_release` as this will ensure compatibility with Ubuntu derivatives :+1:

DavidGoodwin:
so it's not supported for debian variants?

flexiondotorg:
> so it's not supported for debian variants?

Not currently.

DavidGoodwin:
Thanks!

PR #106, opened 2022-05-06
Added alias for `sudo deb-get`

NayamAmarshe:
I am using `echo $0` for telling what's the current shell, and using the name for the rc file. Seems to work fine on ZorinOS but need confirmation from users of other operating systems.

This PR mainly includes changes to the README:
- Adding an alias command in the installation, to alias `deb-get` as `sudo deb-get`.
- Removing all references of sudo that aren't needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ release pages.
 Use `deb-get` to install `deb-get`
 
 ```bash
-sudo apt install curl && curl -sL https://raw.githubusercontent.com/wimpysworld/deb-get/main/deb-get | sudo -E bash -s install deb-get
+sudo apt install curl && curl -sL https://raw.githubusercontent.com/wimpysworld/deb-get/main/deb-get | sudo -E bash -s install deb-get && echo 'alias deb-get="sudo deb-get"' >> ~/.`echo $0`rc
 ```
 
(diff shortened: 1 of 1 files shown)


Tachi107:
I don't think that implicit root permissions are a good thing... If you need to make changes as a super user you should be conscious of that.

Also, are root privileges required for all `deb-get` commands? Like `deb-get list`?

Lastly, not everyone uses `sudo`. There are alternatives like [`doas`](https://wiki.debian.org/Doas)

Tachi107:
Looking at the code it seems that `deb-get` doesn't simply require root privileges unconditionally, but additionally it requires that the user has gained then by calling `sudo`

https://github.com/wimpysworld/deb-get/blob/9024c2d8ce88d83bbb58b23fefa73fe38d078fec/deb-get#L1081-L1089

However, commands like `deb-get list` do not seem to make use of the requested privileges, as they only query some `dpkg` status (I may be wrong here, I only had a brief look).

Instead of `alias`ing `deb-get` to `sudo deb-get`, I think that a cleaner solution would be to only require root privileges when needed (just like `apt` does), and additionally read `DOAS_USER` if `SUDO_USER` is undefined to support `doas` too.

flexiondotorg:
I am not comfortable with obscuring the use of `sudo` in the way. It is my intention to modify `deb-get` so it only checks for root context for actions that require it :slightly_smiling_face:

PR #135, opened 2022-05-10
Allow to run as root user

plattrap:
`sudo` check did not allow to be run as root user.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -1143,7 +1143,9 @@ if [ "$(id -u)" -ne 0 ]; then
   fancy_message fatal "You must use sudo to run $(basename ${0})."
 fi
 
-if [ "${SUDO_USER}" ]; then
+if [ "$(id -u)" -eq "$(id -ru)" ]; then
+  USER_HOME=$(getent passwd "${USER}" | cut -d: -f6)
+elif [ "${SUDO_USER}" ]; then
   USER_HOME=$(getent passwd "${SUDO_USER}" | cut -d: -f6)
 elif [ "${DOAS_USER}" ]; then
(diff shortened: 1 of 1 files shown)


deb-get
```diff
@@ -1143,7 +1143,9 @@ if [ "$(id -u)" -ne 0 ]; then
   fancy_message fatal "You must use sudo to run $(basename ${0})."
 fi
 
-if [ "${SUDO_USER}" ]; then
+if [ "$(id -u)" -eq "$(id -ru)" ]; then
```
flexiondotorg, line 1146: This condition is true for regular users. WHat are you actually trying to test for here?

flexiondotorg (changes requested):
The test you've proposed is also true for regular users in a non-root context.

deb-get
plattrap: If I am running as root already, and not using `sudo` or `doas` then this block tells me to use `sudo` which I do not need to.

Line 1142 checks if my effective user id is `0`, so I would not have gotten here as a normal user.
This checks if my real user id matches my effective user id.
Maybe `if [ "$(id -ru)" -ne 0 ]; then` here would make it clearer?

Captn138: Why not just comparing with `[ "$(id -u)" -eq 0 ]` ?
```suggestion
if [ "$(id -u)" -eq 0 ]; then
```

plattrap: Because if running with `sudo` or `doas` that also shows true. And that is already checked above.

The change is for the case where the command is actually run as the root user. Using `id -ru` to find out who is behind it.

flexiondotorg:
Closed via #186

PR #181, opened 2022-05-22
Only require sudo if mandatory & additional command variations. Close #161

### LINKED ISSUE #161: deb-get --help
Running "sudo deb-get --help" should return the same results as "sudo deb-get" to remain consistent with apt and other command line applications.

Invoking "deb-get --help" should also work without sudo.

torusJKL:
added --help, -h, --version, -v to be consistent with other cli tools.

only request to be run with sudo if it is necessary for the requested action.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -372,6 +372,15 @@ function upgrade_debs() {
     done
 }
 
+function sudo_required()
+{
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Closed via #186.

The addition of `--help` and `--version` is not required. `deb-get` doesn't have options, just commands.

PR #182, opened 2022-05-22
Added LibreOffice Fresh, WhatsDesk, jumpapp

Lupccs:
trying to copy similar solutions on the script, the WhatsDesk one may be a bit tricky but they all seemed to work


flexiondotorg:
Please submit each new package in their own pull requests :+1:

Issue #258, opened 2022-06-23
Add LibreWolf


natanjunges:
- [x] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: LibreWolf
 - Download URL: https://librewolf.net/installation/debian
 - Website URL: https://librewolf.net
 - App Summary: An independent fork of Firefox, with the primary goals of privacy, security and user freedom.

natanjunges:
I'm working on it.

PR #287, opened 2022-07-07
Add LibreWolf

natanjunges:
Implements #258.


flexiondotorg:
I am just having a think about best to handle `CODENAMES_SUPPORTED` for the future. It should be possible to handle via the Debian and Ubuntu codenames only.

PR #326, opened 2022-07-15
Clean should purge the cache directory so that deb-get can update itself

mrm48:
deb-get could not update itself after using deb-get clean (only tried deleting .deb files). This change would ensure .json files are deleted to allow deb-get to update itself when a new release is created.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2453,7 +2453,7 @@ case "${ACTION}" in
         ls -lh "${CACHE_DIR}/";;
     clean)
         elevate_privs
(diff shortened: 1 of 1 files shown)


flexiondotorg (changes requested):
Clean should still delete and `.deb` packages too. Can you reinstate that please :slightly_smiling_face:

mrm48:
:+1: 
Created a new commit to add removal of .deb files back :).

flexiondotorg (approved):

Issue #328, opened 2022-07-15
0.3.1 drops support for Zorin OS


doms9:
### Expected behavior

The package should function normally as it did before version 0.3.1

### Actual behavior

The package refuses to function due to unsupported distro.

### Steps to reproduce

Running any of deb-get's commands would function as intended but after updating to 0.3.1-1 from 0.3.0-1, it returns<br> `[!] ERROR! Zorin OS Focal is not supported because it is not derived from a supported Debian or Ubuntu release.`

#### /etc/os-release

```bash
NAME="Zorin OS"
VERSION="16.1"
ID=zorin
ID_LIKE=ubuntu
PRETTY_NAME="Zorin OS 16.1"
VERSION_ID="16"
HOME_URL="https://zorin.com/os/"
SUPPORT_URL="https://help.zorin.com/"
BUG_REPORT_URL="https://zorin.com/os/feedback/"
PRIVACY_POLICY_URL="https://zorin.com/legal/privacy/"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

#### /etc/lsb-release

```bash
DISTRIB_ID=Zorin
DISTRIB_RELEASE=16
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Zorin OS 16.1"
```

#### deb-get version

```bash
  [!] ERROR! Zorin OS Focal is not supported because it is not derived from a supported Debian or Ubuntu release.
```

### deb-get update

```bash
  [!] ERROR! Zorin OS Focal is not supported because it is not derived from a supported Debian or Ubuntu release.
```

MortenPiil:
Linux Mint 20.3 is also affected, works with 0.3.0

Issue #327, opened 2022-07-15
KDE Neon does not worked on 0.3.1


dtantono:
Expected behavior
The package should function normally as it did before version 0.3.1

Actual behavior
The package refuses to function due to unsupported distro.

Steps to reproduce
Running any of deb-get's commands would function as intended but after updating to 0.3.1-1 from 0.3.0-1, it returns
[!] ERROR! KDE Neon Focal is not supported because it is not derived from a supported Debian or Ubuntu release.

System information
/etc/os-release

  GNU nano 4.8                                                                                                  
NAME="KDE neon"
VERSION="5.25"
ID=neon
ID_LIKE="ubuntu debian"
PRETTY_NAME="KDE neon User - 5.25"
VARIANT="User Edition"
VARIANT_ID=user
VERSION_ID="20.04"
HOME_URL="https://neon.kde.org/"
SUPPORT_URL="https://neon.kde.org/"
BUG_REPORT_URL="https://bugs.kde.org/"
LOGO=start-here-kde-neon
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

/etc/lsb-release
ISTRIB_ID=neon
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="KDE neon User - Plasma 5.25"

deb-get version
  [!] ERROR! KDE Neon  Focal is not supported because it is not derived from a supported Debian or Ubuntu release.

deb-get update
  [!] ERROR! KDE Neon Focal is not supported because it is not derived from a supported Debian or Ubuntu release.

natanjunges:
Please follow the template for reporting bugs. We cannot help without the relevant information.

dtantono:
@natanjunges updated

natanjunges:
OK, looks similar to #328. A solution is being worked in #343.

PR #343, opened 2022-07-16
Support Ubuntu and Debian-derived OS using ID_LIKE in /etc/os-release

onepunchdan:
Tested on Pop!_OS 22.04, this should extend support to Ubuntu and Debian-derived distros that reference either in the ID_LIKE variable of /etc/os-release. Pop!_OS 22.04 in particular lists both `ID_LIKE="ubuntu debian"` in which case only the first is used.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2407,6 +2407,10 @@ esac
 OS_CODENAME=$(lsb_release --codename --short)
 if [ -e /etc/os-release ]; then
     UPSTREAM_ID="$(grep "^ID=" /etc/os-release | cut -d'=' -f2)"
+    UPSTREAM_ID_LIKE="$(grep "^ID_LIKE=" /etc/os-release | cut -d'=' -f2 | sed s/\"//g | cut -d' ' -f1)"
+    if [ ! -z "$UPSTREAM_ID_LIKE" ]; then
+        UPSTREAM_ID=$UPSTREAM_ID_LIKE
+    fi
     case "${UPSTREAM_ID}" in
         debian) UPSTREAM_CODENAME=$(grep DEBIAN_CODENAME /etc/os-release | cut -d'=' -f2);;
         ubuntu) UPSTREAM_CODENAME=$(grep UBUNTU_CODENAME /etc/os-release | cut -d'=' -f2);;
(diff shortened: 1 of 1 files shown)


onepunchdan:
Fixes #328

deb-get
```diff
@@ -2453,7 +2457,7 @@ case "${ACTION}" in
         ls -lh "${CACHE_DIR}/";;
     clean)
         elevate_privs
-        ${ELEVATE} rm -v "${CACHE_DIR}"/*.deb;;
+        ${ELEVATE} rm -v "${CACHE_DIR}"/*.deb
```
natanjunges, line 2460: Please refrain from changing things outside of the scope of your pull request. It helps keep things organized.

natanjunges (changes requested):

deb-get
onepunchdan: Ah, my apologies. I should have made a separate commit for that but I believe there's now another PR with the same correction.

dominix (commented):
works for me on pop!_os 22.04

philclifford:
Confirmed working on Linux Mint
closes #352

philclifford:
may not work for #327 because neon has

```
ID_LIKE="ubuntu debian"
```

ymauray:
This broke `deb-get` 0.3.2, at least on Ubuntu MATE 22.04. `/etc/os-release` has 
```
ID=ubuntu
ID_LIKE=debian
```

The code provided by this PR then replaces `UPSTREAM_ID` with `debian`, which breaks all packages that install from a PPA, given they all look like this : 
```shell
function deb_copyq() {
    if [ "${UPSTREAM_ID}" == "ubuntu" ]; then
        PPA="ppa:hluk/copyq"
        PRETTY_NAME="CopyQ"
        WEBSITE="https://hluk.github.io/CopyQ/"
        SUMMARY="An advanced clipboard manager with editing and scripting features."
    fi
}
```
making it impossible to update or install packages from PPAs : 
```shell
sudo deb-get install copyq
  [!] ERROR! Download URL for copyq is empty. Skipping.
```

Issue #331, opened 2022-07-16
Strawberry install problem


Xmetalfanx:
### Expected behavior

Describe the expected behavior here.

Tried to install Stawberry Media Player via `deb-get install strawberry `

### Actual behavior

Describe the actual behavior here.

Seeing 

```
 ~  deb-get install strawberry                                                                                                                          6.3s  Sat 16 Jul 2022 03:58:48 AM EDT
ERROR: invalid ppa name 'strawberry '
  [!] ERROR! Failed to add: ppa:jonaski/strawberry . Deleting it.
ERROR: invalid ppa name 'strawberry '

```

### Steps to reproduce

Describe the exact steps required to reproduce the issue.

Run `deb-get install Strawberry`

### System information

Run the following a paste the output below:

#### /etc/os-release
```
PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

#### /etc/lsb-release
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04 LTS"
```

#### deb-get version
```
deb-get version                             
0.3.1
```

### deb-get update
```
deb-get update                            
Hit:1 http://us.archive.ubuntu.com/ubuntu jammy InRelease
Hit:2 http://us.archive.ubuntu.com/ubuntu jammy-updates InRelease
Hit:3 http://us.archive.ubuntu.com/ubuntu jammy-backports InRelease
Hit:4 http://security.ubuntu.com/ubuntu jammy-security InRelease
Hit:5 https://packages.microsoft.com/repos/vscode stable InRelease
Ign:6 https://repo.vivaldi.com/stable/deb stable InRelease
Hit:7 https://repo.vivaldi.com/stable/deb stable Release
[...]

PR #346, opened 2022-07-16
Strawberry PPA code update

Xmetalfanx:
- Fix for https://github.com/wimpysworld/deb-get/issues/331


Xmetalfanx:
I haven't been able to check that this actually fixes the issue but I am guessing the code isn't handling that space correctly ... none of the other PPA functions seem to have that space and i HAVE installed that PPA before and even after i opened that issue ... that is the correct PPA

PR #349, opened 2022-07-17
Added : output list of available software in CSV format

ymauray:
I'm working on a thing that needs the list of available software in a machine readable format, so I create this function to output the list of available softwares in CSV format.
I have tested the output with that thing I'm working on, and it works fine.
I'm available if changes need to be implemented, as those could impact my thing.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -356,6 +356,13 @@ function prettylist_debs() {
     done
 }
 
+function csvlist_debs() {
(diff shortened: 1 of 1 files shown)


natanjunges:
The README should also be updated, preferably with [cog](https://pypi.org/project/cogapp).

ymauray:
Darn it ! I always forget the README.md !! Thanks. PR updated.

PR #357, opened 2022-07-18
getting debian working

philclifford:
There is no DEBIAN_ in `/etc/os-release` on debian bullseye
Both Ubuntu and Debian seem to use VERSION_CODENAME


philclifford:
Runs but fails to work on Debian 11
 - failing to populate cache - more to sort out

PR #366, opened 2022-07-20
Protect Ubuntus from derivatives fix

philclifford:
This has been tested on Ubuntu Mate, POP, Mint, MX, KDE Neon and Debian and fixes ppa installs on Ubuntu (devcontainer) and Ubuntu Mate and tested ubuntu derivatives
closes #365

PR #373, opened 2022-07-20
Try source of apt-add-repository to determine appropriate parameters

### LINKED ISSUE #370: Linux Mint has no option in apt-add-repository
### Expected behavior

PPAs should be added to Linux Mint using `apt-add-repository -y --no-update "${PPA}"`.

### Actual behavior

As the `apt-add-repository` command in Linux Mint is actually a rewrite in Python of the original tool from Debian, it does not support the `--no-update` option.

### Steps to reproduce

Try to install any package from a PPA in Linux Mint:
```shell
sudo deb-get install <pkg>
```

### Possible solution
It seems that the effect of the `--no-update` option is the default in the Linux Mint implementation, so this option could be removed when Linux Mint is detected.

philclifford:
Only use `--no-update`parameter for `apt-add-repository` if it comes from software-properties-common
hopefully a working and viable solution rather than checking and parsing parameters

works on Ubuntu Mate (with #366 in place)
also tested on

- [x] Mint
- [x] Pop
- [x] KDE neon
- [x] Debian
- [x] MX Linux xfce (no ppa support -  ID=debian)

closes #370

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -181,7 +181,12 @@ function install_apt() {
 }
 
 function install_ppa() {
-    if ${ELEVATE} apt-add-repository -y --no-update "${PPA}"; then
+    local AAR_PARAMS
+    AAR_PARAMS="-y"
+    if [ $(dpkg -S $(command -v apt-add-repository)|cut -d: -f1) == "software-properties-common" ] ; then
+        AAR_PARAMS+=" --no-update";
+    fi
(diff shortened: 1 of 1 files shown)


philclifford:
@flexiondotorg on Debian the `list` actions list "ppa" applications (with blank publication info and so if you try and install a ppa computer says no which is probably for the best in most cases ) . Would it be better not to `list` apps that debian users cannot `deb-get install` ? (Or perhaps hide them behind a `--show-unsupported`switch ? )

PR #378, opened 2022-07-22
Add tests and return non zero exit code when failed to install package

OhMyMndy:
In this branch I added the following

- Return a non zero status code when a package cannot be installed
- Added some rudimentary tests on Github Actions, which shows right away that `firefox-esr` and `obs-studio` cannot be installed ([!] ERROR! Download URL for firefox-esr is empty. Skipping.)

Might be interesting to have a look at this pull request during one of your streams in conjunction with configuring [branch protection rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)?

For example, adding a branch protection rule which prevents merging to `main` when the `Tests` workflow doesn't run successfully to have this as quality control.

[Terraform could be used to configure the branch protection rules for example](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/branch_protection)


philclifford:
Great minds ... having been playing with devcontainers and looking at github actions (because `bin-get` used them :-) Thanks! ),  I've built a small monster that tries to do tests on both ubuntu and debian. Lots of fun and learning.  Yours will be far better
but I'll still enjoy improving mine.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,53 @@
...
+          # Dependencies to build the deb
+          # sudo apt-get install build-essential devscripts debhelper -y
+
+          # Dependencies for deb-get itself
+          sudo apt-get install apt apt-transport-https bash coreutils curl dpkg grep libc-bin lsb-release sed software-properties-common sudo wget -y
+
+
+      # All these commands should return a 0 status
+      - name: Test zero status
+        run: |
```
philclifford, line 28: need to run update first to create /var/cache/deb-get (or ...)
else upstream/main currently fails
see https://github.com/philclifford/deb-get/runs/7484384462?check_suite_focus=true#step:4:8
but https://github.com/philclifford/deb-get/runs/7484487001?check_suite_focus=true looks good

.github/workflows/tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Test package installations
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          set -x
+          # @todo fix firefox-esr see https://github.com/OhMyMndy/deb-get/runs/7467951397?check_suite_focus=true
+          ./deb-get install firefox-esr || true
+          ./deb-get install 1password >/dev/null
+          ./deb-get install bitwarden >/dev/null
+          # @todo fix obs-studio see https://github.com/OhMyMndy/deb-get/runs/7467973508?check_suite_focus=true
+          ./deb-get install obs-studio  || true
```
philclifford, line 44: firefox-esr and obs-studio now work on upstream/main because there was a bug that is no more.
Thanks for the loan of your code and the awesome learning day I've had today.

PR #387, opened 2022-07-24
Properly handle PPAs on distros that are not derived from Ubuntu

### LINKED ISSUE #386: prettylist, csvlist and show do not list unsupported PPAs
### Expected behavior

As `prettylist` is used to generate the `README` file with `cog`, it should always list all packages. Other commands such as `csvlist` and `show` should also always list the information of all packages.

### Actual behavior

When running on distros that are not derived from Ubuntu, the packages that use PPAs do not export their detail info, and the outputs of `prettylist`, `csvlist` and `show` are incomplete. This also makes the detection of the installation method fallback to `website`, which could introduce further unexpected behaviour.

### Steps to reproduce

On any distro not derived from Ubuntu, run:
```shell
deb-get prettylist
```

natanjunges:
Fixes #386/#379, is cleaner, more modular, less prone to errors and provides better error message.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -425,7 +425,7 @@ function upgrade_debs() {
     for APP in "${APPS[@]}"; do
         # Only upgrade .debs that are installed
         if package_is_installed "${APP}"; then
-            validate_deb "${APP}" 1
+            validate_deb "${APP}"
             if [ "${METHOD}" == "direct" ] || [ "${METHOD}" == "github" ] || [ "${METHOD}" == "website" ]; then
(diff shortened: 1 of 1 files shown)


M4rQu1Nh0S:
This pull #387 fixes the issue.

Tested in Linux Mint 20.3 & Ubuntu 22.04.

flexiondotorg:
Very nice!

philclifford:
[Tests on The Matrix](https://github.com/philclifford/deb-get/actions/workflows/testingmatrix.yml)
:point_up: on `ubuntu-latest, debian-latest,linuxmint,mxlinux, popos, zorin, kdeneon`

PR #396, opened 2022-07-27
Add VirtualBox 6.1

### LINKED ISSUE #385: Add VirtualBox
## Requesting a new app ✨

 - App Name: `VirtualBox`
 - Download URL: https://www.virtualbox.org/wiki/Linux_Downloads
 - Website URL: https://www.virtualbox.org/
 - App Summary: `VirtualBox is a powerful x86 and AMD64/Intel64 virtualization product for enterprise as well as home use.`

EDIT: they are offering different .deb files for different versions of Ubuntu, I don't know if this is a problem though.

philclifford:
Working [here](https://github.com/philclifford/deb-get/runs/7571438547?check_suite_focus=true)

Closes #385

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -532,18 +532,16 @@ function deb_quickemu() {
     SUMMARY="Quickly create and run optimised Windows, macOS and Linux desktop virtual machines."
 }
 
-function deb_virtualbox() {
+function deb_virtualbox-6.1() {
     ARCHS_SUPPORTED="amd64"
-    if [ ! -e  /usr/share/keyrings/oracle-virtualbox-2016.gpg ] ; then
-        curl -sL https://www.virtualbox.org/download/oracle_vbox_2016.asc | sudo gpg --dearmor --yes --output /usr/share/keyrings/oracle-virtualbox-2016.gpg
-    fi
     APT_KEY_URL="https://www.virtualbox.org/download/oracle_vbox_2016.asc"
-    APT_LIST_NAME="virtualbox"
-    APT_REPO_URL="deb [arch=${HOST_ARCH} signed-by=/usr/share/keyrings/oracle-virtualbox-2016.gpg] https://download.virtualbox.org/virtualbox/debian ${OS_CODENAME} contrib"
-    PRETTY_NAME="VirtualBox"
+    APT_LIST_NAME="virtualbox-6.1"
+    APT_REPO_URL="deb [arch=amd64] https://download.virtualbox.org/virtualbox/debian ${OS_CODENAME} contrib"
+    PRETTY_NAME="VirtualBox 6.1"
     WEBSITE="https://www.virtualbox.org/"
-    SUMMARY="VirtualBox is a general-purpose full virtualizer for x86 hardware, targeted at server, desktop and embedded use."
+    SUMMARY="VirtualBox 6.1 is a general-purpose full virtualizer for x86 hardware, targeted at server, desktop and embedded use."
 }
+
 
 function deb_firefox-esr() {
     PPA="ppa:mozillateam/ppa"
--- a/deb-get
+++ b/deb-get
@@ -533,18 +533,17 @@ function deb_quickemu() {
 }
 
 function deb_virtualbox() {
-    if [ ! -e  /usr/share/keyrings/oracle-virtualbox-2016.gp ] ; then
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -532,6 +532,19 @@ function deb_quickemu() {
     SUMMARY="Quickly create and run optimised Windows, macOS and Linux desktop virtual machines."
 }
 
+function deb_virtualbox() {
+    ARCHS_SUPPORTED="amd64"
+    if [ ! -e  /usr/share/keyrings/oracle-virtualbox-2016.gpg ] ; then
```
natanjunges, line 537: This should only run when installing, and a counterpart should be provided when removing. Ideally, this should not be required at all.

natanjunges (changes requested):

deb-get
philclifford: Without it I was getting failures, it only runs when the key isn't there and the inbuilt  removal code already removes it I think (or the deb uninstall maybe).
Will look further as I may have used this to fix a typo I also already fixed or some such,,,,, it is very late(early) 

[Tests](https://github.com/philclifford/deb-get/actions/workflows/tests.yml)

[here is the run](https://github.com/philclifford/deb-get/runs/7532056726?check_suite_focus=true#step:5:16)

philclifford:
```
Err:15 https://download.virtualbox.org/virtualbox/debian jammy InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A2F683C52980AECF
Fetched 154 kB in 5s (31.8 kB/s)
Reading package lists...
[...]
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://download.virtualbox.org/virtualbox/debian jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A2F683C52980AECF
W: Failed to fetch https://download.virtualbox.org/virtualbox/debian/dists/jammy/InRelease  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A2F683C52980AECF
W: Some index files failed to download. They have been ignored, or old ones used instead.
Selecting previously unselected package virtualbox.
```
which goes on to  install on my system because I have "old ones" and tonight (ahem - this 3am actually) ...
So I think best to leave it there and sleep on it
Cheers

philclifford: So,  now I remember what a pain Virtualbox upstream could be, and why I generally avoided the "debian" packaging.
Trying to rename the app to `virtualbox-6.1` so you actually get that not whatever ubuntu is providing in `virtualbox` ...
well now I broke it so it is marked draft while I go quietly mad(der)

natanjunges: But is this the actual name of the package? Because if not, it would break the detection of the package being installed.

philclifford:
Yes, they have `virtualbox-6.0` and `virtualbox-5.2` in the same repo too.
Making the list name just `virtualbox` might be the lesser of two evils,
but as this stands it now works and actually installs (and recognises) the
upstream version rather than picking whatever is in the distro repository.

deb-get
philclifford: Yes, they have `virtualbox-6.0` and `virtualbox-5.2` in the same repo too.
Making the list name just `virtualbox` might be the lesser of two evils,
but as this stands it now works and actually installs (and recognises) the
upstream version rather than picking whatever is in the distro repository.

deb-get
```diff
@@ -532,6 +532,17 @@ function deb_quickemu() {
     SUMMARY="Quickly create and run optimised Windows, macOS and Linux desktop virtual machines."
 }
 
+function deb_virtualbox-6.1() {
+    ARCHS_SUPPORTED="amd64"
+    APT_KEY_URL="https://www.virtualbox.org/download/oracle_vbox_2016.asc"
+    APT_LIST_NAME="virtualbox-6.1"
+    APT_REPO_URL="deb [arch=amd64] https://download.virtualbox.org/virtualbox/debian ${OS_CODENAME} contrib"
```
natanjunges, line 539: I'm pretty sure your are supposed to use `UPSTREAM_CODENAME` instead.

natanjunges (changes requested):

deb-get
philclifford: I did, but was having failures with some alien distros, Looking at it again now.

PR #400, opened 2022-07-27
Add Nala

GHMusicalCoder:
Nala is an apt replacement that handles parallel Downloads and easier ability to visualize what is happening with installs and updates


GHMusicalCoder:
Closing this pull request - this is not working right - didn't realize I was getting a GPG error in testing...

PR #404, opened 2022-07-29
Case sensitivity causing issues

kz6fittycent:
Pop!_OS capitalizes "Jammy" and deb-get won't continue. Hoping this fixes the issue. Could be causing issues on other distros.

In case https://github.com/wimpysworld/deb-get/commit/fe4cb91c60535c438af685894a4093e58eb241d4 doesn't fix it.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 LC_ALL=C
 
-readonly VERSION="0.3.3"
+readonly VERSION="0.3.4" #version not verified
 
 function usage() {
 cat <<HELP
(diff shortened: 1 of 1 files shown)


deb-get
```diff
@@ -2505,14 +2505,14 @@ else
 fi
 
 case "${UPSTREAM_CODENAME}" in
```
flexiondotorg, line 2507: Would be better to just test for a lowercased codename using `${UPSTREAM_CODENAME,,}` rather the regex everything.

flexiondotorg (changes requested):

deb-get
kz6fittycent: Agreed. Closing PR.

natanjunges:
Why did you close the PR? You just had to fix what was requested.

Issue #44, opened 2022-05-05
Add ivpn


dnet890:
ivpn is a open source VPN application
They have a deb available from their website.
https://repo.ivpn.net/stable/pool/ivpn_3.8.7_amd64.deb

flexiondotorg:
Download available from:
 - https://github.com/ivpn/desktop-app/releases
 - https://www.ivpn.net/apps-linux/#ubuntu

PR #414, opened 2022-08-02
Added IVPN for Desktop

Xmetalfanx:
Closes https://github.com/wimpysworld/deb-get/issues/44

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2413,6 +2413,17 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_ivpn() {
+    get_github_releases "https://api.github.com/repos/ivpn/desktop-app/releases/latest"
+    if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -2413,6 +2413,17 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_ivpn() {
+    get_github_releases "https://api.github.com/repos/ivpn/desktop-app/releases/latest"
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL=$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | cut -d'"' -f4)
```
natanjunges, line 2419: This does not work, as the URL is in the body of the release.

natanjunges (changes requested):

Xmetalfanx:
I am taking this PR down .... not sure i like different options to get the URL since everything i try can get messed up if the maintainer puts things in a different order (say the two DEBs and RPMs in a different order than they are currently)

natanjunges:
Analysing the download URLs, it seems that they have a fixed pattern where the only thing that changes is the version number, so you could just get the version information from the GitHub release and use it to define the `URL` variable.

Issue #411, opened 2022-07-30
Add Peek Gif recorder


FXBOY4EVA:
## Requesting a new app ✨

 - [✓ ] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: Peek
 - Download URL: https://code.launchpad.net/~peek-developers/+archive/ubuntu/stable
   (It is available in Debian repositories, as well as via AppImage, Flatpak and other distributions' repositories).
 - Website URL: https://github.com/phw/peek
 - App Summary: Peek - Simple animated GIF screen recorder with an easy to use interface

Peek has been one of my go-to tools for a while.
I figured since it has a PPA run by the developers it should be available via deb-get. Also figured Martin might enjoy this one.

flexiondotorg:
Thanks for working on this. However, the versions of peek published in the Ubuntu archive since 20.04 are newer than the versions of peek published in the PPA, so I will not be adding this change.

PR #415, opened 2022-08-02
Added Peek - an animated GIF recorder

Xmetalfanx:
Closes https://github.com/wimpysworld/deb-get/issues/411


flexiondotorg:
Thanks for working on this. However, the versions of peek published in the Ubuntu archive since 20.04 are newer than the versions of peek published in the PPA, so I will not be adding this change.

Xmetalfanx:
I did not see that.   No problem ... I figured with some of these PR's I'd try to address a few of the listed "issues" to slim that list of issues down.

Issue #89, opened 2022-05-05
Add Master-PDF-Editor


dtantono:
Master PDF Editor powerful PDF Editor
They have a deb available from their website.
https://code-industry.net/public/master-pdf-editor-5.8.52-qt5.x86_64.deb

PR #418, opened 2022-08-02
Added Master PDF Editor

Xmetalfanx:
Closes https://github.com/wimpysworld/deb-get/issues/89

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2413,6 +2413,16 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_master-pdf-editor() {
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL=$(curl -s -S "https://code-industry.net/get-master-pdf-editor-for-ubuntu/?download" | grep -Eo "href=\"https://code-industry.net/public/.*\.deb" | cut --complement -c -6)
+        VERSION_PUBLISHED="$(echo "${URL}" | cut -d'/' -f6)"
+    fi
+    PRETTY_NAME="Master PDF Editor"
+    WEBSITE="https://code-industry.net/free-pdf-editor/"
+    SUMMARY="Edit PDF files in Linux. It enables you to create, edit, view and print interactive PDF documents."
+}
+
 # Create an array to track those deb_ functions being loaded by this script
 readonly DEB_GET_APPS=($(declare -F | grep deb_ | sed 's|declare -f deb_||g' | sort))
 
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -2413,6 +2413,16 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_master-pdf-editor() {
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL=$(curl -s -S "https://code-industry.net/get-master-pdf-editor-for-ubuntu/?download" | grep -Eo "href=\"https://code-industry.net/public/.*\.deb" | cut --complement -c -6)
+        VERSION_PUBLISHED="$(echo "${URL}" | cut -d'/' -f6)"
```
natanjunges, line 2419: Should this not be `cut -d'-' -f5`?

natanjunges (changes requested):

deb-get
```diff
@@ -2413,6 +2413,16 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_master-pdf-editor() {
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL=$(curl -s -S "https://code-industry.net/get-master-pdf-editor-for-ubuntu/?download" | grep -Eo "href=\"https://code-industry.net/public/.*\.deb" | cut --complement -c -6)
```
natanjunges, line 2418: Multiple files are provided, depending on the version of Debian/Ubuntu. The correct file should be selected based on `UPSTREAM_CODENAME`, and the supported codenames should be listed in `CODENAMES_SUPPORTED`. Read [CONTRIBUTING#creating-the-function](https://github.com/wimpysworld/deb-get/blob/main/CONTRIBUTING.md#creating-the-function) for further information.

natanjunges (changes requested):

deb-get
natanjunges: There is one for Debian 8 - 10 / Ubuntu 14.x - 20.x and another for Debian 9 - 11 / Ubuntu 18.x - 22.x.

natanjunges:
Why did you close it?

deb-get
Fuseteam: one seems to include qt the other doesn't, there's one download button on the page
also 18.x - 22.x should cover everything from ubuntu to 18.x 22.x so that should include ubuntu 20.x while debian 9-11 should cover everything from debian 9 to 11, so that should include debian 10

deb-get doesn't support anything older than ubuntu anyway so the default download link should cover everything we need

Xmetalfanx:
sorry for the lack of response (dealing with Hurricane Ian and  Ian recovery here lately) ... if i am not mistaken after a few alterations I just figured that I didn't want to have a PR submitted if it "wasn't quiet ready yet" ... I do see some else has a version of that that maybe better with a second commit they did.  I just felt like having PR's with more than a few changes needed on my end is not ready and i didn't want to "clutter" up the PR list (so to speak) ... that was my reasoning

Issue #164, opened 2022-05-13
Add Xournal++


dnet890:
notetaking app
https://xournalpp.github.io/
They have a deb available from their website.
https://github.com/xournalpp/xournalpp/releases/tag/v1.1.1

natanjunges:
I'll work on it.

natanjunges:
Turns out that this package is already in the Debian and Ubuntu repositories.

Xmetalfanx:
Oops ... wish I would have scrolled down before doing that commit (no PR yet) .... I had the same question about another issue (knowing the versions in say LTS releases and the latest the dev offers MAY not be the same, at some points ) regarding adding Bleachbit.

If the versions go out of sync and I am not talking a few days for it to make it to the Ubuntu repo .... that may be a logical reason to add different requests ... but the "core deb-get team" would have to answer that logic i suppose.   I know software coming from the Ubuntu repos will not be accepted, I think it was

natanjunges:
@flexiondotorg I understand by reading [CONTRIBUTING#what-is-acceptable](https://github.com/wimpysworld/deb-get/blob/main/CONTRIBUTING.md#what-is-acceptable) that those packages would be rejected. What is your saying in this?

Xmetalfanx:
was in the process of revisiting this one and then seeing it is in the Debian and Ubuntu repos that it'd be rejected on those grounds .. maybe this issue should be closed if that is the case (to start tidying up the list of issues)

soni17:
It's already available in ubuntu and debian repositories under the name "xournalpp". I suggest we close this.

PR #421, opened 2022-08-03
Added Xournal++

Xmetalfanx:
Closes https://github.com/wimpysworld/deb-get/issues/164

Tested/Working locally

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2414,11 +2414,14 @@ function deb_mullvad-vpn() {
 }
 
 function deb_xournalpp() {
-    CODENAMES_SUPPORTED="bionic bullseye focal"
+    CODENAMES_SUPPORTED="bullseye focal"
     get_github_releases "https://api.github.com/repos/xournalpp/xournalpp/releases/latest"
-    if [ "${ACTION}" != "prettylist" ]; then 
+    if [ "${ACTION}" != "prettylist" ]; then
         URL="$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)"
         VERSION_PUBLISHED="$(echo "${URL}" | cut -d'/' -f8)"
+        echo -e "URL:\t${URL}"
+        echo -e "VERSION_PUBLISHED:\t:${VERSION_PUBLISHED}"
+        exit
     fi
     PRETTY_NAME="Xournal++"
     WEBSITE="https://github.com/xournalpp/xournalpp/"
--- a/deb-get
+++ b/deb-get
@@ -2413,6 +2413,20 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_xournalpp() {
+    CODENAMES_SUPPORTED="bionic bullseye focal"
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -2413,6 +2413,17 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_xournalpp() {
+    get_github_releases "https://api.github.com/repos/xournalpp/xournalpp/releases/latest"
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL="$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)"
```
natanjunges, line 2419: Multiple files are provided, depending on the version of Debian/Ubuntu. The correct file should be selected based on `UPSTREAM_CODENAME`, and the supported codenames should be listed in `CODENAMES_SUPPORTED`. Read [CONTRIBUTING#creating-the-function](https://github.com/wimpysworld/deb-get/blob/main/CONTRIBUTING.md#creating-the-function) for further information.

natanjunges (changes requested):

deb-get
```diff
@@ -2413,6 +2413,20 @@ function deb_mullvad-vpn() {
     SUMMARY="VPN Service based in Sweden that does not collect user data."
 }
 
+function deb_xournalpp() {
+    CODENAMES_SUPPORTED="bionic bullseye focal"
```
natanjunges, line 2417: `bionic` is not supported by `deb-get`.

deb-get
```diff
@@ -2413,6 +2413,20 @@ function deb_mullvad-vpn() {
...
+    get_github_releases "https://api.github.com/repos/xournalpp/xournalpp/releases/latest"
+    if [ "${ACTION}" != "prettylist" ]; then
+        if [[ "${UPSTREAM_CODENAME}" == *"${CODENAMES_SUPPORTED}"* ]]; then
```
natanjunges, line 2420: There is no need for this check, as it is already done before this function is called.

deb-get
```diff
@@ -2413,6 +2413,20 @@ function deb_mullvad-vpn() {
...
+            URL="$(grep "browser_download_url.*.deb" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)"
```
natanjunges, line 2421: To `grep` the correct URL, you could do something like `"browser_download_url.*${UPSTREAM_CODENAME}-x86_64\.deb\""`.

natanjunges (changes requested):

Xmetalfanx:
I am pulling down this PR until i get a chance to fix the issues and squah commits into one when approriate

deb-get
Xmetalfanx: thank you for all the suggestions on how to improve my code additions ... I addressed a few of these issues (pulled down some PR's for now, since they clearly aren't ready from my end, i realize that now) 

this issue i have hopefully addressed but i need to get a bullseye or focal VM loaded to test it ... was wondering why my latest test code wasn't returning a URL but i am on a Jammy base and this app doesn't support Jammy (last I checked)

PR #441, opened 2022-08-13
fix mergerfs saying there's an update when there's not.

dhitchcock:
Fixed the VERSION_PUBLISHED of mergerfs to match the format of VERSION_INSTALLED so it's not constantly saying there's an update when there's not.


flexiondotorg (approved):
LGTM

PR #449, opened 2022-08-15
Use PAT to get higher github rate-limit

philclifford:
Goes from 60/hour to 5000/hour for authenticated sessions.

This should hopefully avoid serious `deb-get`ters (and those testing repeatedly :raised_hands: ) from running into rate-limiting issues too early and often.


flexiondotorg:
Thanks for working on this. Please can you add a few words to README.md about how to user this feature?

philclifford:
I have added a section to the README.  I also spent a day trying to work out why the test workflow is failing (including when essentially the same PR previously passed. (see https://github.com/philclifford/deb-get/actions/runs/2871542877 and friends) 
Some of that was what triggered me to work on this (testing failing because all the tests all come from one/very few sources unless you use the provided secret token) .  Found some other things that could do with tweaking along the way, which I'll put in another issue, but could not get their `ubuntu-latest` (?focal !!) runner to pass. Maybe a chat on discord with @OhMyMndy will bear fruit there.

philclifford:
Somewhere along the line I broke it - I fried my brain with the quoting and inserting
```
 '--header "Authorization: token $DEBGET_TOKEN"'
```
into the wget to the api only if set and without  the whole lump rejected as an unknown parameter .... resorting to the array seemed to fix it but now It looks like I've been using old cache entries from previous tests/installs and silent wgets have masked the fact that they're not running at all ....

echoing the array and running it by hand works a treat so it is close, but no ${\\"\"\\"'cigar}

philclifford:
replaced by #457

PR #450, opened 2022-08-17
Install correct versions of rstudio for jammy/focal

LyzardKing:
Rstudio has two versions: one for jammy and one for bionic/focal, with the only difference being the libssl version it depends on.
This PR adds a case statement to pick the correct version


flexiondotorg (approved):
LGTM

PR #469, opened 2022-08-26
support for Ubuntu devel (rolling)

laniusone:
I know that current devel (kinetic) is technically supported, but when installing it via Rolling Rhino Remix ISO, resulting codename is 'devel' and not 'kinetic', so the script doesn't recognize the distribution. I added a tiny fix to the case statement.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2607,7 +2607,7 @@ case "${UPSTREAM_CODENAME}" in
     focal)    UPSTREAM_RELEASE="20.04";;
     impish)   UPSTREAM_RELEASE="21.10";;
     jammy)    UPSTREAM_RELEASE="22.04";;
-    kinetic)  UPSTREAM_RELEASE="22.10";;
+    kinetic|devel)  UPSTREAM_RELEASE="22.10";;
     *) fancy_message fatal "${OS_ID_PRETTY} ${OS_CODENAME^} is not supported because it is not derived from a supported Debian or Ubuntu release.";;
 esac
 


natanjunges:
Please provide the output of:
```shell
cat /etc/lsb-release
cat /etc/os-release
```

laniusone:
```
/bin/cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.10
DISTRIB_CODENAME=devel
DISTRIB_DESCRIPTION="Ubuntu Kinetic Kudu (development branch)"
```

```
/bin/cat /etc/os-release
PRETTY_NAME="Rolling Rhino Remix"
NAME="Ubuntu"
VERSION_ID="22.10"
VERSION="22.10 (Kinetic Kudu)"
VERSION_CODENAME=devel
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.rollingrhino.org"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=devel
LOGO=ubuntu-logo
```

natanjunges:
This distribution really does not follow the standards from neither Debian nor Ubuntu. It could be argued if we should even bother supporting those distributions, since the whole purpose of using those standards is not having to deal with each case individually. But as the required changes are not so big, I guess it is fine.

laniusone:
Ok, thank you! I will create an issue in their repo anyway, so that they fix it, because I guess it should follow the standards, as technically it is just the devel branch of kinetic with some extra tools.

natanjunges:
Actually, thinking about the implications of the change you propose, unless the distribution changes its `/etc/os-release` file, it might be impractical to support it. `UPSTREAM_CODENAME` should only be the codename of a release of either Debian or Ubuntu, since the functions for the packages rely on that. And the `UBUNTU_CODENAME` field in `/etc/os-release` should only be the codename of a release of Ubuntu (hence the name), and `VERSION_CODENAME` should be used for custom codenames such as this one.

laniusone:
I actually agree that this should be fixed on the distro-side, so I'm closing this.

PR #484, opened 2022-09-02
Fix REDIR_URL for deadbeef

pxc:
The original link wasn't finding anything for me. This uses the travis CI RSS feed, which contains the links.


natanjunges:
This issue was also reported in #495, and a solution that fixes the original link instead of changing it to something else is proposed at #498.

pxc:
Thanks @natanjunges. I retract this in favour of #498.

PR #490, opened 2022-09-03
Added deb_vagrant entry

chriswayg:
I tested this on Ubuntu  `jammy` `amd64`, but it should work on many versions of Debian and Ubuntu as it is written in Ruby. I checked that the repo works with `buster bullseye focal impish jammy`. I am not certain about additional architectures, since the Vagrant website does not specifically mention that.


flexiondotorg (changes requested):
Please remove the change that switches the fatal message to a warning. Please also remove impish, it is EOL.

PR #492, opened 2022-09-04
Kali support

ikeman32:
Added support for Kali in the OS_ID section on line 2580 just below Zorin.

At line 2603 assigns "focal" to UPSTREAM_CODENAME if the OS_ID_PRETTY is "Kali GNU/Linux Rolling" otherwise UPSTREAM_CODENAME remains empty.

Tested install from:
curl -sL https://raw.githubusercontent.com/ikeman32/deb-get/KaliSupport/deb-get | sudo -E bash -s install deb-get

And installs to a Kali VM without issue.

Successful installs of Obsidian and VS Codium so same VM.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2577,7 +2577,7 @@ case "${OS_ID}" in
   Pop) OS_ID_PRETTY="Pop!_OS";;
   Ubuntu) OS_ID_PRETTY="Ubuntu";;
   Zorin) OS_ID_PRETTY="Zorin OS";;
-  #Kali) OS_ID_PRETTY="Kali GNU/Linux Rolling"
+  Kali) OS_ID_PRETTY="Kali GNU/Linux Rolling";;
   *) fancy_message fatal "${OS_ID} is not supported.";;
 esac
 
--- a/deb-get
+++ b/deb-get
@@ -2577,6 +2577,7 @@ case "${OS_ID}" in
   Pop) OS_ID_PRETTY="Pop!_OS";;
   Ubuntu) OS_ID_PRETTY="Ubuntu";;
   Zorin) OS_ID_PRETTY="Zorin OS";;
+  #Kali) OS_ID_PRETTY="Kali GNU/Linux Rolling"
   *) fancy_message fatal "${OS_ID} is not supported.";;
 esac
 
@@ -2608,7 +2609,13 @@ case "${UPSTREAM_CODENAME}" in
     impish)   UPSTREAM_RELEASE="21.10";;
     jammy)    UPSTREAM_RELEASE="22.04";;
     kinetic)  UPSTREAM_RELEASE="22.10";;
-    *) fancy_message fatal "${OS_ID_PRETTY} ${OS_CODENAME^} is not supported because it is not derived from a supported Debian or Ubuntu release.";;
+    *) 
+        if [ "$OS_ID_PRETTY" == "Kali GNU/Linux Rolling" ]; then
+            return 1
+        else
+        
+            fancy_message fatal "${OS_ID_PRETTY} ${OS_CODENAME^} is not supported because it is not derived from a supported Debian or Ubuntu release."
+        fi
 esac
 
 if [ -n "${1}" ]; then
--- a/deb-get
+++ b/deb-get
@@ -2613,7 +2613,6 @@ case "${UPSTREAM_CODENAME}" in
         if [ "$OS_ID_PRETTY" == "Kali GNU/Linux Rolling" ]; then
             return 1
         else
-        
             fancy_message fatal "${OS_ID_PRETTY} ${OS_CODENAME^} is not supported because it is not derived from a supported Debian or Ubuntu release."
         fi
 esac
--- a/deb-get
+++ b/deb-get
@@ -2577,7 +2577,7 @@ case "${OS_ID}" in
   Pop) OS_ID_PRETTY="Pop!_OS";;
   Ubuntu) OS_ID_PRETTY="Ubuntu";;
   Zorin) OS_ID_PRETTY="Zorin OS";;
-  Kali) OS_ID_PRETTY="Kali GNU/Linux Rolling";;
+  Kali) OS_ID_PRETTY="Kali GNU/Linux Rolling";;# add Kali support
   *) fancy_message fatal "${OS_ID} is not supported.";;
 esac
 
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -2595,6 +2596,12 @@ if [ -e /etc/os-release ]; then
         debian) UPSTREAM_CODENAME=$(grep VERSION_CODENAME /etc/os-release | cut -d'=' -f2);;
         *) UPSTREAM_CODENAME="";;
     esac
+
+    # Set UPSTREAM_CODENAME to focal OS_ID_PRETTY is Kali GNU/Linux Rolling
+    # To emulate a Debian derived release
+    if [ "${OS_ID_PRETTY}" == "Kali GNU/Linux Rolling" ]; then
```
natanjunges, line 2602: This is not how UPSTREAM_CODENAME should be handled. If the current solution for detecting UPSTREAM_CODENAME does not work for Kali, then a better way should be proposed.

natanjunges (changes requested):

ikeman32:
All conflicts resolved and refactoring complete. Is working on kali VM. Unsure why tests are failing line 1: says permission denied and exits with code 128.

`
Run ./deb-get update >/dev/null
  ./deb-get update >/dev/null
  ./deb-get list >/dev/null
  ./deb-get help >/dev/null
  ./deb-get cache >/dev/null
  ./deb-get version >/dev/null
  shell: /usr/bin/bash -e {0}
/home/runner/work/_temp/b[2](https://github.com/wimpysworld/deb-get/actions/runs/3094286415/jobs/5007500916#step:4:2)456450-f[3](https://github.com/wimpysworld/deb-get/actions/runs/3094286415/jobs/5007500916#step:4:3)a8-[4](https://github.com/wimpysworld/deb-get/actions/runs/3094286415/jobs/5007500916#step:4:4)272-a9a4-f49b11b[5](https://github.com/wimpysworld/deb-get/actions/runs/3094286415/jobs/5007500916#step:4:5)d1db.sh: line 1: ./deb-get: Permission denied
Error: Process completed with exit code 12[6](https://github.com/wimpysworld/deb-get/actions/runs/3094286415/jobs/5007500916#step:4:6).
`

natanjunges:
Your implementation is not ideal, as `UPSTREAM_ID` should always be either `debian` or `ubuntu`, and `UPSTREAM_CODENAME` should always be the codename of a release of one of the two. This is so that `deb-get` never needs to care about which actual distribution it is running on, just which Debian/Ubuntu release the distribution is compatible with, reducing the need to handle each edge case individually (such as this one). Can you please provide the contents of `/etc/os-release`? This could help justify/clarify your implementation.

ikeman32:
This is the /etc/os-release for the Kali VM:

PRETTY_NAME="Kali GNU/Linux Rolling"
NAME="Kali GNU/Linux"
ID=kali
VERSION="2022.3"
VERSION_ID="2022.3"
VERSION_CODENAME="kali-rolling"
ID_LIKE=debian
ANSI_COLOR="1;31"
HOME_URL="https://www.kali.org/"
SUPPORT_URL="https://forums.kali.org/"
BUG_REPORT_URL="https://bugs.kali.org/"

As you can see there is no DEBIAN_CODENAME but is based on Debian as shown by ID_LIKE. The if blocks that read the os_release file and determine if a release is Ubuntu or Debian derivative does recognize kali-rolling as being derived from either. Kali for some inexplicable reason has extra characters that cannot be seen so they have to be stripped out. 

Here's a look at /etc/os-release on the host machine for the Kali VM:

NAME="Linux Mint"
VERSION="20 (Ulyana)"
ID=linuxmint
ID_LIKE=ubuntu
PRETTY_NAME="Linux Mint 20"
VERSION_ID="20"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.linuxmint.com/"
BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
PRIVACY_POLICY_URL="https://www.linuxmint.com/"
VERSION_CODENAME=ulyana
UBUNTU_CODENAME=focal

Which is quite different from Kali. There may be a better implementation and perhaps someone with better coding skills than my own can come up with one. This implementation though, perhaps not ideal is the only one that I have been able to see. Perhaps a fresh pair of eyes may be able to see what I cannot, but Kali seems to be an unforeseen edge case.

ikeman32:
/etc/debian_version on the Kali VM
reads simply
kali-rolling

I have re-examined my implementation I found that keeping the Fallback check as I have implemented:
`if [ "${UPSTREAM_ID}" != kali ] && [ "${UPSTREAM_ID}" != ubuntu ] && [ "${UPSTREAM_ID}" != debian ]; then`

I was able to remove the if else block that checks for Kali and sets the UPSTREAM_CODENAME and leave it as it was before. But still requires a check for the UPSTREAM_ID after all the other checks to set the UPSTREAM_CODENAME.

Place after the Debian 12 + check on line 
`
# Kali
if [ "${UPSTREAM_ID}" = kali ]; then
    UPSTREAM_CODENAME=$(grep "^VERSION_CODENAME=" ${OS_RELEASE} | cut -d'=' -f2 | sed s/\"//g | cut -d' ' -f1)   
fi
`
Or in order to remove Kali from the fall back and eliminate the second UPSTREAM_ID check after the Debian 12 + check then the check immediately above would have to be changed from:

`UPSTREAM_CODENAME=$(grep "^VERSION_CODENAME=" ${OS_RELEASE} | cut -d'=' -f2)`

to: ` UPSTREAM_CODENAME=$(grep "^VERSION_CODENAME=" ${OS_RELEASE} | cut -d'=' -f2 | sed s/\"//g | cut -d' ' -f1)`

Because the Kali /etc/os_release has extra spaces/characters that must be stripped out or the kali-rolling pattern in the case below will not be triggered and throw an ERROR. I don't have or know of any other distro where this edge case applies so I can't test if the change would break on distros without an Ubuntu/Debian UPSTREAM_CODENAME. Although I don't know why it would.

I'm happy to push which ever option you think is best, though I'm thinking the second option is probably the preferred one.

natanjunges (approved):

natanjunges (changes requested):
Please rebase with main.

ikeman32:
After clearing conflicts fixed minor typo.

takov751:
I just saw this and kali-rolling is based on debian testing according to their documentation https://www.kali.org/docs/policy/kali-linux-relationship-with-debian/

So far all i had to add to main deb-get `kali`  to `OS_ID` and add  `DEBIAN_CODENAME=bookworm` to /etc/os-release.
I know that this is not ideal, but this way I am making sure that none of the installs breaks

deb-get
```diff
@@ -2685,7 +2686,8 @@ else
     fancy_message fatal "os-release not found. Quitting"
 fi
 
-UPSTREAM_ID="$(grep "^ID=" ${OS_RELEASE} | cut -d'=' -f2)"
+
+UPSTREAM_ID="$(grep "^ID=" /etc/os-release | cut -d'=' -f2)"
```
philclifford, line 2690: Please revert this line change

philclifford (changes requested):

PR #514, opened 2022-09-10
re-added Vagrant

chriswayg:
Re-submitting this pull request after re-syncing with upstream and making the [requested changes](https://github.com/wimpysworld/deb-get/pull/490) to the _vagrant_ section. I tested this on Ubuntu `jammy` `amd64`, but it should work on many versions of Debian and Ubuntu as it is written in Ruby. I checked that the repo works with buster bullseye focal and jammy. I am not certain about additional architectures, since the Vagrant website does not specifically mention that.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -3,7 +3,7 @@ LC_ALL=C
 PACKAGE_INSTALLATION_TRIES=0
 PACKAGE_INSTALLATION_COUNT=0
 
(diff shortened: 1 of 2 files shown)


flexiondotorg (changes requested):
Please rebase with `main`.

PR #521, opened 2022-09-13
allow AppimageLauncher to be installed in Debian

### LINKED ISSUE #520: cant install appimagelauncher
i cant install appimagelauncher by using deb-get 

it yells about thats its not ubuntu ppas but if i go to the releases of the appimage launcher download it

install it

works just fine thats kinda stoopid.

silentJET85:
fixes #520 

This provides a way to install AppimageLauncher in Debian, while continuing to use the PPA in Ubuntu.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -2481,7 +2481,12 @@ function deb_papirus-icon-theme() {
 }
 
 function deb_appimagelauncher() {
-    PPA="ppa:appimagelauncher-team/stable"
+    ARCHS_SUPPORTED="amd64 i386 armhf arm64"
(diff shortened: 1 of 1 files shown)


natanjunges:
A package should not have more than one installation method, since it would produce different outputs to `prettylist` in different systems. I think that in this case the best solution would be to replace the PPA with the GitHub releases method in all cases.

flexiondotorg (changes requested):
Please consolidate this change so Ubuntu and Debian install via the same mechanism.

silentJET85:
done

flexiondotorg (approved):

Issue #465, opened 2022-08-23
Shell completion


yermulnik:
## Requesting a new feature 🧑‍💻

Would be great to have Shell completion (for Bash at least) bundled into a `deb-get` package.
Is there an item in `deb-get`'s roadmap on adding such a feature?

### Expected behavior

Shell completion is installed along with `deb-get` package and users can utilise tab-completion for commands and parameters.

### Actual behavior

Shell completion is not install and hasn't been added to `deb-get` as of yet.

### Steps to reproduce

Install `deb-get` and run `deb-get<tab><tab>`

### System information

Run the following a paste the output below:

#### /etc/os-release

```bash
> cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal```

#### /etc/lsb-release

```bash
> cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.4 LTS"```

#### deb-get version

```bash
> deb-get version
0.3.5
```

natanjunges:
I like this idea, anything that can save me a few key presses.

natanjunges:
I came up with an implementation, but without the optimizations from https://github.com/natanjunges/deb-get/tree/extrepo, which is a WIP implementation for #313, it is too slow. I'll wait until #313 is finished and merged before submitting it.

yermulnik:
@natanjunges Thank you for the time and effort 🙇🏻

natanjunges:
Closed by #523.

Issue #313, opened 2022-07-12
Support external repositories and move supported packages to one


natanjunges:
**This is a long-term proposal**.

The list of supported packages is growing very fast, and hardcoding it into the main script does not scale well:
- Loading metadata of all supported packages when only a few are installed could inflate the package size and degrade performance;
- Changes in packages are only available in the next release;
- Changes in packages get mixed with changes in the actual script;
- As mentioned in #269, the `README` gets polluted with the list of supported packages.

To address those issues, I propose adding support to external repositories, and moving all supported packages to one. A repository would consist of:
- A manifest file listing all supported packages;
- A script file for each supported package.

When a repository is added, only the manifest file is downloaded. The script for a certain package would only be downloaded when required. It would also only be loaded in the main script when required, as the manifest can be used to get the list of supported packages. The wrapping `deb_` functions could even be removed from the scripts.

flexiondotorg:
I agree that we need to separate the package definitions from the main script 👍 Making a monolithic script was a conscious choice at the very beginning, in order to quickly bootstrap the project. 

Also, I did not anticipate just how many packages would be available to include in `deb-get` 😁

I'd like to make one exception, and that is for `deb-get` itself. I want `deb-get` to be able to install/update itself without needing to query an *"external"* manifest.

I would also like to get Debian and Codename support in `deb-get` before we work on this

Fuseteam:
i was just thinking filing an issue about this xd

natanjunges:
Nice, if you have some suggestions, feel free to share.

Fuseteam:
i do have an idea but i think i need to work it out first

natanjunges:
@flexiondotorg an implementation is available at https://github.com/natanjunges/deb-get/tree/extrepo, but I'll wait for all other pending PR's to be merged before opening this one, so all conflicts are sorted out. Also, the repository for the main external repository should be created first.

peutch:
+1 for this

And should deb-get project be the place to collect any external repositories like for: https://packages.icinga.com/?
:-)

peutch:
On Debian Wiki there is a list: https://wiki.debian.org/DebianRepository/Unofficial

flexiondotorg:
@natanjunges I think it is time to move forward with this.

Trying to merge the ever-growing list of new packages into the monolithic script is no longer maintainable :sob: 

[...]

Issue #295, opened 2022-07-07
geforcenow-electron dropped support for deb packages


natanjunges:
### Expected behavior

The package should be installed normally.

### Actual behavior

The developers announced in the latest release notes that the support for deb packages was dropped, so the release does not have a deb package any more:
```
  [+] Updating /var/cache/deb-get/geforcenow-electron.json
  [!] ERROR! Download URL for geforcenow-electron is empty. Skipping.
```

### Steps to reproduce

Run:
```bash
sudo deb-get install geforcenow-electron
```

#### /etc/os-release

```
PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

#### /etc/lsb-release

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04 LTS"
```

#### deb-get version

```
0.3.0
```

### deb-get update

```
Hit:1 https://download.docker.com/linux/ubuntu jammy InRelease
Ign:2 http://packages.linuxmint.com una InRelease
Get:3 https://zotero.retorque.re/file/apt-package-archive ./ InRelease [2153 B]
Hit:4 http://archive.ubuntu.com/ubuntu jammy InRelease
Hit:5 http://packages.linuxmint.com una Release
Get:6 http://archive.ubuntu.com/ubuntu jammy-updates InRelease [114 kB]
Get:9 http://archive.ubuntu.com/ubuntu jammy-backports InRelease [99.8 kB]
Hit:8 https://packagecloud.io/AtomEditor/atom/any any InRelease
Get:10 http://archive.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
Fetched 326 kB in 2s (150 kB/s)
Reading package lists...
  [+] Updating /var/cache/deb-get/deb-get.json
```

[...]

PR #523, opened 2022-09-16
Add support to external repositories, bash completion and other quality of life improvements

natanjunges:
Implemented features:
- [x] External package repositories (implements #313)
- [x] Bash completion (implements #465)
- [x] Deprecated packages handling, warning about them when installed (mentioned at #295)
- [x] Optimized and filtered versions of the `list` command (`--raw`, `--installed` and `--not-installed`)
- [x] Filter outputs of `prettylist` and `csvlist` by package repository (reverts and fixes #458)
- [x] Simpler version of the `update` command (`--repos-only`) that only initializes and updates `deb-get`'s external package repositories, without updating apt or looking for updates of installed packages
- [x] Install/upgrade hooks to initialize/update package repositories
- [x] Purge hook to remove cache and package repositories
- [x] Documentation of new features and preparation of existing pertinent documentation to be moved to main package repository
- [x] Split test workflows that apply to `deb-get` and to `01-main` subfolder


philclifford:
I used this (or something very like it) to move some custom user includes across:

```bash
#!/bin/bash
DGLIDESTPATH=/etc/deb-get/99-local.d
for file in /etc/deb-get.d/*; do
    tst=$(grep function "${file}")
    fn=$(basename "${file}")
    if [ "${tst/${fn}/}" == "function deb_() {" ]; then
    #the file is named exactly as the app name
        tail +2 "${file}" | grep -v -e '^}$' -e '^$' >   "${DGLIDESTPATH}/${fn}"
    else
        echo "# ERROR: File ${file} does not conform for app ${fn}"
        tst2="${tst/() {}"
        echo "# it appears to be for the app : ${tst2/function deb_}"
    fi
done
```

rough and ready and presumptive, but worked.  
A parser to pull out the main `deb_<app>() ` functions into a repository is a little more challenging, or have you got one already?

natanjunges:
No, I created each file individually in my test repository. They are not many (186), but a script could definitely help.

philclifford:
Definitely worth automating in case the base script gains and loses some apps.  I'll try and dig out something from the dark corners of my past lives (sure there are some self-modifying recursive nightmares lurking there but probably in C or AREXX or PROLOG ;-)  ... probably quicker to google something from stackoverflow with awk ...  
I suppose someone will be wanting `fish` completions too ...  ;-)

natanjunges:
@philclifford or @OhMyMndy, can one of you please work in the task "Prepare pertinent test workflows to be moved to main package repository"? What is needed is to split the test workflow in two, one only with the tests of `deb-get` itself, and the other with the tests of installation of the packages. You can fork my fork and open PRs to it, so I can integrate them into this one.

philclifford:
Happy to try. Would be good to discuss/plan in a wiki or maybe via discord when time permits

diddledani (approved):
I'm gonna say LGTM.

philclifford:
I still believe https://github.com/wimpysworld/deb-get/pull/523/commits/d08c9435efd855db79b4b9ab960f0aef82dd7d40 breaks us (because 2 APT list pointing to same repo with two (albeit identical) keys makes apt upset - testing once VMs are ready.

Otherwise LGTM too - bravo!

I'll try and let all the install tests run through somewhere (also on debian) but it is noteworthy that
 * the "deb-get Tests" passes even though dpkg fails if you look at the log :shrug:
 * several apps will now fail on kinetic because repos/ppas have not built for kinetic (yet)
   *  currently that causes the list and key to be removed - maybe not ideal if there was a happy jammy repo

natanjunges:
> I still believe [d08c943](https://github.com/wimpysworld/deb-get/commit/d08c9435efd855db79b4b9ab960f0aef82dd7d40) breaks us

Yes indeed, but I thought it was better to address that in a separate PR once this one is merged (I'm thinking on a solution with hard links, since they have built-in reference counting, but there might be better ones). Then we can work on the ideas we discussed for `apt`-like `depends`, `conflicts` and `replaces` as well.

>     * the "deb-get Tests" passes even though dpkg fails if you look at the log shrug

Where? I only found an `ls` error, but it is harmless (ideally it should be silenced).

>     * several apps will now fail on kinetic because repos/ppas have not built for kinetic (yet)

This does not concern this PR, but we will indeed need to address it, ideally before the next release of `deb-get`.

>       * currently that causes the list and key to be removed - maybe not ideal if there was a happy jammy repo

We should also try to come up with a better solution in a separate PR, something in the lines of #603.

philclifford:
> Yes indeed, but I thought it was better to address that in a separate PR once this one is merged

Fine with that : we should perhaps add a note somewhere that 6.1 users must uninstall 6.1 before installing 7.0

philclifford:
> Where? I only found an `ls` error, but it is harmless (ideally it should be silenced).

Sorry - mea culpa.  Looks like I was confused (how unusual!)  - Only seeing the `ls` now: I did see a dpkg fail in a test that passed, but was thrashing around in the early hours and mis-remembered after sleeping.

philclifford (approved):
https://github.com/philclifford/deb-get/actions/runs/3331912181
[deb-get Tests](https://github.com/philclifford/deb-get/actions/workflows/tests-deb-get.yml)

LGTM

OhMyMndy:
Really cool to see some big progress on `deb-get`! One thing, which isn't possible any more, is to just grab the `deb-get` file from the repo and execute it because it depends on many other files. Reason why I used `deb-get` like that is to use the version from my pull request because it wasn't merged yet (I wanted to install something with `deb-get` which was not upstream yet).

This setup resolves that problem by allowing external repo's!

Great work @natanjunges @philclifford @diddledani

flexiondotorg (approved):

Issue #528, opened 2022-09-18
Keys moved to the wrong place


fethomm:
# deb-get issue template

Thank you for taking the time to report an issue to the `deb-get` project 🙂

 - Please complete one the appropriate sections below.
 - Delete all the sections that do not apply, including this one.
 - Any issues that do not use this template will be closed.
 - If you're not sure what to do, or have a general question about `deb-get`, then ask in [Wimpy's World Discord](https://wimpysworld.io/discord).

## Requesting a new app ✨

If you are requesting a new app to be added to `deb-get` please complete the
following section, only one app per issue please.

 - [ ] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name:
 - Download URL:
 - Website URL:
 - App Summary:

## Requesting a new feature 🧑‍💻

If you are proposing a new feature for `deb-get` please describe the proposed
feature with as much detail as possible here:

## Reporting a bug 🐛

Please make sure the version of `deb-get` you are using is up to date:

```bash
deb-get update
deb-get reinstall deb-get
```

### Expected behavior

Describe the expected behavior here.
Keys of installed apps should go to /usr/share/keyrings
### Actual behavior

Describe the actual behavior here.
Keys of installed apps go to deprecated /etc/apt/trusted.gpg or etc/apt/trusted.gpg.d
### Steps to reproduce
deb-get install foo and check where the key gets copied to
Describe the exact steps required to reproduce the issue.

### System information

Run the following a paste the output below:

#### /etc/os-release
Ubuntu 22.04
```bash
cat /etc/os-release


#### /etc/lsb-release

```bash
cat /etc/lsb-release
```

#### deb-get version

```bash
deb-get version
0.3.5
### deb-get update

```bash
deb-get update
```

takov751:
After looking into it . Install deb function works as it's expected and adds key to keyring directory, however apt-add-repository is used in add_ppa function, which causing this issue as it's the old way still. Might need to be implement manually adding ppa to source list and keyring

takov751:
Until there is a proper replacement. I was wondering if we should implement in such way from ppa:username:ppa_name just get the username and ppa_name as variable. Download key and move it to keyring then create file into sources.list.d 
```echo "deb [arch=amd64 signed-by=/usr/share/keyrings/R.gpg] https://cloud.r-project.org/bin/linux/ubuntu jammy-cran40/" | sudo tee /etc/apt/sources.list.d/R.list > /dev/null```
something like this

[...]

PR #534, opened 2022-09-21
Move all apt keys to /usr/share/keyrings

natanjunges:
Implements #528. As keys stored in `/etc/apt/trusted.gpg.d` are applied globally, they are unsafe, and doing so will soon be deprecated. This implementation dearmors all ASCII GPG keys and stores them in `/usr/share/keyrings`, only signing their respective repositories. It also works with PPAs, treating them as regular apt repositories. It however changes slightly how packages are declared.


flexiondotorg (approved):
Fabulous! Thank you <3

Forage:
...and with that none of the repos show up in the "Other Software" of the "Software & Updates" app any longer...

PR #463, opened 2022-08-21
Added Bashtop/Bpytop PPA

kz6fittycent:

PR #541, opened 2022-09-23
Removed Bashtop PPA, added Wraith-Master entry

kz6fittycent:
The Bashtop PPA is unwarranted and I've removed it, closed PR #463 

This new PR includes Wraith-Master. Built `deb-get` with added changes locally and tested with success. 

Addresses: #542


natanjunges (changes requested):
Please remove all changes in the `README` file that do not concern your package.

kz6fittycent:
I just ran cog -r README.md per the contribution instructions. So do I just need to manually edit ask that out?

natanjunges:
Ideally no, since every new package added would add their own entry in `README`. But some of them didn't, so you need indeed to manually remove their entries from `README` (a PR to update those other entries is already being proposed).

kz6fittycent:
I have an idea to perhaps mitigate this in terms of workflow - what if the README was updated once just prior to release? That way all changes are final (per the upstream team) and no "go-backs"?

natanjunges:
I meant all *changes*, not *all* entries. The idea of a once-per-release `README` update is good.

kz6fittycent:
Resync'd from your main - ran `cog -r README.md` that provides Wraith-Master as the only add. I hope this gets it, otherwise, I'm afraid I'm at a loss.

natanjunges (approved):

PR #544, opened 2022-09-24
Fix logic error in validate_deb

natanjunges:


kz6fittycent:
`cog -r README.md` now works as intended.

deb-get
```diff
@@ -357,7 +357,7 @@ function validate_deb() {
     elif  [ -n "${PPA}" ]; then
         METHOD="ppa"
     else
-        if (([ "${METHOD}" = github ] && [ -e "${CACHE_DIR}/${APP}.json" ]) || ([ "${METHOD}" != github ] && [ "${ACTION}" != "prettylist" ])) && ([ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]); then
+        if (([ "${METHOD}" = github ] && [ -e "${CACHE_DIR}/${APP}.json" ]) || [ "${METHOD}" != github ]) && [ "${ACTION}" != "prettylist" ] && ([ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]); then
```
philclifford, line 360: It's 3am and my mother tongue is SQL so I might be confused or old-fashioned, but I'm pretty sure you should use "{ ..; }"  parenthesis to group those tests rather than "(..)" if only for performance reasons.
https://wiki.bash-hackers.org/commands/classictest#precedence_and_parenthesis

philclifford: Can confirm this fixes `prettylist` - thus cog of README.
Need more time and coffee to apply brain to logic without crashing.

natanjunges: > I'm pretty sure you should use "{ ..; }" parenthesis to group those tests rather than "(..)"

Indeed, I did not connect the fact that the parentheses I used to mark precedence are actually sub-shells.

flexiondotorg (approved):
LGTM

PR #546, opened 2022-09-25
fix: Extend test coverage on basic commands

### LINKED ISSUE #545: Workflow test misses failure of prettylist command
The cog stanza uses prettylist and that is not tested in the github workflow so if it breaks ....

philclifford:
to catch breakage of more commands

fixes #545


flexiondotorg (approved):
LGTM

Issue #556, opened 2022-09-29
Add Cider


kz6fittycent:
## Requesting a new app ✨

 - [x] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: Cider 
 - Download URL: https://github.com/ciderapp/cider-releases/releases/download/v1.5.6/cider_1.5.6_amd64.deb
 - Website URL: https://github.com/ciderapp/Cider
 - App Summary: A new cross-platform Apple Music experience based on Electron and Vue.js written from scratch with performance & visuals in mind.

PR #557, opened 2022-09-29
Add Cider

kz6fittycent:
See #556

Adds Cider to `deb-get`. Running successfully on local machine. Updated README.


kz6fittycent:
Accidentally closed...sorry about that.

kz6fittycent:
My fault - should've created a new branch in my repo to PR VBOX 7.0. My apologies again. This PR can/will address #556 and #576 respectively. I'll update the PR title.

deb-get
```diff
@@ -633,15 +613,15 @@ function deb_quickemu() {
     SUMMARY="Quickly create and run optimised Windows, macOS and Linux desktop virtual machines."
 }
 
-function deb_virtualbox-6.1() {
```
natanjunges, line 636: Since the name of the package for VirtualBox was changed, it is not that simple to upgrade it. Please remove the changes made to it, so we can accept the Cider package.

README.md
```diff
@@ -58,62 +58,45 @@ help = result.replace("Usage\n\n", "").rstrip()
 cog.out(f"```\n{help}\n```")
 ]]] -->
 ```
-
```
natanjunges, line 61: Please revert those changes that do not concern the Cider package.

deb-get
```diff
@@ -16,67 +16,47 @@ fi
 
 function usage() {
 cat <<HELP
-
```
natanjunges, line 19: Please revert those changes that do not concern the Cider package.

natanjunges (changes requested):

deb-get
```diff
@@ -806,6 +786,17 @@ function deb_wraith-master() {
     SUMMARY="An application for controlling the RGB LEDs on AMD's Wraith stock coolers"
 }
 
+function deb_cider() {
+    get_github_releases "https://api.github.com/repos/ciderapp/cider-releases/releases/latest"
+    if [ "${ACTION}" != "prettylist" ]; then
+        URL=$(grep "browser_download_url.*\.deb\"" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)
+        VERSION_PUBLISHED="$(echo "${URL}" | cut -d'/' -f8 | sed 's|release-||')"
```
ngilles, line 793:
```suggestion
        VERSION_PUBLISHED="$(echo "${URL}" | cut -d'/' -f8 | sed 's|^v||')"
```

ngilles (changes requested):

deb-get
kz6fittycent: I made the change, but it seems like it's fine either way.

kz6fittycent: I see. I made the change per your request. Thanks for clarifying.

ngilles: for completeness sake, this is the "issue":

The URL that's extract is like this:
`https://github.com/ciderapp/cider-releases/releases/download/v1.5.7/cider_1.5.7_amd64.deb`

Then to extract the version it picks:
The `cut` command in there splits it at the `/`s, (`-d'/'`)and get's the 8th field (`-f8`):
`https:` `<empty>` `github.com`  `ciderapp` `cider-releases` `releases` `download` `v1.5.7` `cider_1.5.7_amd64.deb`
The 8th field being `v1.5.7`

Which to get the actual version number is then fed through to the `sed` command which replaces `^v` (start of string and character `v`) with nothing (empty string), yielding `1.5.7`

The original code that you copied from was trying to remove `release-` (likely because that project was using `release-x.y.z`), so that final step would not break per se, but would leave the `v` in there in front of the version number.

kz6fittycent:
Reverted Vbox changes. Just Cider now.

natanjunges:
There are still all the changes in both the `README` and `deb-get`.

README.md
kz6fittycent: I've removed these lines - didn't see them, even though I scrolled to the line(s) in question. Probably human error. I've removed them. I think this is a good case for my proposal to only update the README once, just prior to publishing a new revision. It might've mitigated the go-backs. That said, in this case, it was my error but this might happen again for someone else.

natanjunges:
The format of package definitions has changed. Please read [01-main/CONTRIBUTING](https://github.com/wimpysworld/deb-get/blob/main/01-main/CONTRIBUTING.md) and [EXTREPO](https://github.com/wimpysworld/deb-get/blob/main/EXTREPO.md), update your implementation and submit a new PR.

PR #592, opened 2022-10-16
fake gitkraken for unpublised version

philclifford:
Avoid repeated re-installs of 8.9.1 because they have pulled 8.10.0 but left it in the release list

Draft because we don't really want to merge this but it seems to do the job


flexiondotorg:
Issue fixed upstream.

PR #593, opened 2022-10-16
add new non-ppa candidate, even though I nabbed it first

philclifford:


philclifford:
They've nixed it so no point in this

PR #602, opened 2022-10-23
Fix breakage from updated system-monitoring-center URL

### LINKED ISSUE #601: system-monitoring-center broke us
The URL lost a few  "_" so we lost the released version.

philclifford:
better solution needed, but this works
fixes #601


philclifford:
An alternative might be to
```
VERSION_PUBLISHED="$(echo "${URL}" |  grep -o '/v.*/'|tr -d '/v' )"
```

but that is just as breakable, maybe

```
VERSION_PUBLISHED="$(echo "${URL}" | grep -o -e '[0-9]\.[0-9][0-9]\.[0-9]' )"
```

or

```
VERSION_PUBLISHED="$(echo "${URL}" | grep -o -E '[[:digit:]]\.[[:digit:]]+\.[[:digit:]]' )"
```

is better.

natanjunges:
Fixed by #523.

Issue #548, opened 2022-09-25
Add LibreOffice Fresh


kingeejit:
- I have checked that this app does not already have a pending request and has not previously been declined.
  - Previous (declined) request in #453 was for the tar zipped deb direct from libreoffice.org rather than the Fresh PPA published by Ubuntu's Libreoffice packaging team.

- App Name: libreoffice
- Download URL: ppa:libreoffice/ppa
- Website URL: https://www.libreoffice.org/
- App Summary: LibreOffice Fresh is the latest (non-alpha/beta) release of the newest series of LibreOffice — a free and powerful office suite.

kingeejit:
When the original request came in, there had been a delay in releasing the 7.4 series to the PPA. This is no longer the case and the PPA is tracking the website closely.

I've not noticed any significant delays previously, so I think the PPA makes a worthwhile substitute for the website and is more easily consumed.

natanjunges:
Closed by #620.

PR #620, opened 2022-10-30
Add LibreOffice Fresh

mikabak:
Requested here #548


01-main/packages/libreoffice
```diff
@@ -0,0 +1,6 @@
+DEFVER=1
+PPA="ppa:libreoffice/ppa"
+EULA=""
```
natanjunges, line 3: No need to add an empty variable.

natanjunges (changes requested):

natanjunges (approved):

Issue #537, opened 2022-09-22
Add Git Credential Manager


RayJW:
## Requesting a new app ✨

- [x] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: Git Credential Manager
 - Download URL: https://github.com/GitCredentialManager/git-credential-manager/releases/latest
 - Website URL: https://github.com/GitCredentialManager/git-credential-manager/
 - App Summary:

[Git Credential Manager](https://github.com/GitCredentialManager/git-credential-manager) (GCM) is a secure Git credential helper built on [.NET](https://dotnet.microsoft.com/) that runs on Windows, macOS, and Linux.

Compared to Git's [built-in credential helpers](https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage) (Windows: wincred, macOS: osxkeychain, Linux: gnome-keyring/libsecret) which provides single-factor authentication support working on any HTTP-enabled Git repository, GCM provides multi-factor authentication support for [Azure DevOps](https://dev.azure.com/), Azure DevOps Server (formerly Team Foundation Server), GitHub, Bitbucket, and GitLab.

Git Credential Manager (GCM) replaces the .NET Framework-based [Git Credential Manager for Windows](https://github.com/microsoft/Git-Credential-Manager-for-Windows) (GCM), and the Java-based [Git Credential Manager for Mac and Linux](https://github.com/microsoft/Git-Credential-Manager-for-Mac-and-Linux) (Java GCM), providing a consistent authentication experience across all platforms.

natanjunges:
Closed by #621.

PR #621, opened 2022-10-30
Add Git Credential Manager

mikabak:
Requested here #537


01-main/packages/gcm
```diff
@@ -0,0 +1,11 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64"
```
natanjunges, line 2: This is already the default, so not needed.

natanjunges (changes requested):

natanjunges (approved):

PR #622, opened 2022-10-30
fix examples in EXTREPO.md

mikabak:
EXTREPO.md: Fixing syntax error in examples


natanjunges:
None of those are syntax errors, they are perfectly valid bash syntax.

PR #627, opened 2022-10-30
Fix flameshot to use jammy release until kinetic is built

philclifford:
Seems to work fine

image

And also should now work on debian rather than breaking because the old pattern hard-coded 'ubuntu'

### DIFF
--- a/01-main/packages/flameshot
+++ b/01-main/packages/flameshot
@@ -1,8 +1,12 @@
-DEFVER=3
+DEFVER=1
 CODENAMES_SUPPORTED="buster bullseye focal jammy kinetic"
 get_github_releases "https://api.github.com/repos/flameshot-org/flameshot/releases/latest"
 if [ "${ACTION}" != "prettylist" ]; then
-    URL="$(grep  -E "browser_download_url.*\.${UPSTREAM_ID}-${UPSTREAM_RELEASE:0:2}.*\.${HOST_ARCH}\.deb\"" "${CACHE_DIR}/${APP}.json" | tail -n1 | cut -d'"' -f4)"
+    case ${UPSTREAM_RELEASE} in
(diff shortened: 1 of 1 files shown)


01-main/packages/flameshot
```diff
@@ -1,9 +1,14 @@
...
 if [ "${ACTION}" != "prettylist" ]; then
-    URL="$(grep "browser_download_url.*\.${UPSTREAM_ID}-${UPSTREAM_RELEASE}\.${HOST_ARCH}\.deb\"" "${CACHE_DIR}/${APP}.json" | head -n1 | cut -d'"' -f4)"
+    case ${UPSTREAM_RELEASE} in
+        22.10) ONLY_ONE="tail -1" ;;
+        *) ONLY_ONE="head -1"
+    esac
+    URL="$(grep  -E "browser_download_url.*\.${UPSTREAM_ID}-${UPSTREAM_RELEASE:0:2}.*\.${HOST_ARCH}\.deb\"" "${CACHE_DIR}/${APP}.json" | $ONLY_ONE | cut -d'"' -f4)"
     local VERSION_TMP="${URL##*/flameshot-}"
-    VERSION_PUBLISHED="${VERSION_TMP%%[-.]ubuntu*}"
+    VERSION_PUBLISHED="${VERSION_TMP%%[-.]${UPSTREAM_ID}*}"
```
philclifford, line 11: At least this needed to enable debian support

01-main/packages/flameshot
```diff
@@ -1,9 +1,14 @@
-DEFVER=1
+DEFVER=3
```
natanjunges, line 1: This shouldn't be incremented.

natanjunges (changes requested):

natanjunges (approved):

flexiondotorg (approved):

PR #628, opened 2022-10-30
Fix OpenRGB version and URL extraction

philclifford:
More upstream twiddling handled


01-main/packages/openrgb
```diff
@@ -1,16 +1,16 @@
-DEFVER=1
+DEFVER=2
```
natanjunges, line 1: This shouldn't be incremented.

natanjunges (changes requested):

natanjunges (approved):

flexiondotorg (changes requested):
Please rebase with main

flexiondotorg (approved):

PR #629, opened 2022-10-30
Improve debian support after tests

philclifford:
mainly just marking the uninstallable ones 
Giving up chasing this until #636 and #635 land, at which point this will need rebasing or have been obsoleted.


philclifford:
re-worked in #678

PR #633, opened 2022-10-31
pick correct beersmith version for release

philclifford:
They don't make it easy. This semi-hard-coding seems necessary for older releases.
Installs and works without issue on kinetic and now should pick the recommended versions for focal and bionic.
[01-main Tests](https://github.com/philclifford/deb-get/actions/workflows/tests-01-main.yml)

### DIFF
--- a/01-main/packages/beersmith3
+++ b/01-main/packages/beersmith3
@@ -1,6 +1,21 @@
 DEFVER=1
+CODENAMES_SUPPORTED="bionic focal jammy kinetic"
 if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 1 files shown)


philclifford:
Still tries and fails to install on debian, in spite of restricting to ubuntu codenames :shrug: :crying_cat_face:

01-main/packages/beersmith3
```diff
@@ -1,6 +1,21 @@
-DEFVER=1
+DEFVER=3
```
natanjunges, line 1: This shouldn't be incremented.

natanjunges (changes requested):

flexiondotorg (changes requested):
Please rebase with main

flexiondotorg (approved):

PR #635, opened 2022-10-31
Filter out unsupported packages in list

natanjunges:
For now the performance is really bad, but #636 should fix it.


philclifford:
Running well (so long as broken/changed upstreams are handled via fixes in `99-local` ( or adding 
`sudo echo "https://raw.githubusercontent.com/philclifford/deb-get/extras/91-fixes"  > /etc/deb-get/91-fixes.repo`
)

flexiondotorg (approved):

PR #640, opened 2022-11-02
RStudio Desktop fixed for move to posit.co

### LINKED ISSUE #638: RStudio has moved
Now the website redirects [here](https://posit.co/)  and all the download info has been "improved" we no longer find the URL.

Working on a solution, but also considering adding `rstudio-server` ( #641 )  and maybe even `r-base` 

NB testing hampered because local overrides were broken by 4dc0c33 ( see #639 ) , but 
`git checkout d274c36 ; ./deb-get show rstudio` suggests a workable fix will be along shortly.

philclifford:
Also updated to point debians at the recommended package.

closes #638

### DIFF
--- a/01-main/packages/rstudio
+++ b/01-main/packages/rstudio
@@ -1,5 +1,4 @@
-DEFVER=1
-get_website "https://www.rstudio.com/products/rstudio/download/"
+DEFVER=2
 if [ "${ACTION}" != "prettylist" ]; then
     case "${UPSTREAM_CODENAME}" in
         focal|buster|bullseye)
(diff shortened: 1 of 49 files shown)


01-main/packages/rstudio
```diff
@@ -1,15 +1,16 @@
-DEFVER=1
+DEFVER=2
```
natanjunges, line 1: This shouldn't be incremented.

natanjunges (changes requested):

natanjunges (approved):

flexiondotorg (changes requested):
Please rebase with main

philclifford:
Need to re-sub

PR #672, opened 2022-11-17
deprecate balena-etcher-electron

philclifford:
We have a choice to leave this enabled, have users install or upgrade to a version that doesn't run at all and tell them to wait (optionally also to install the AppImage and run it with "--no-sandbox" ) until a working deb upgrade materializes or to deprecate it until it is fixed upstream. My preference is the latter, hence this PR

Raised upstream, where they just showed a list of all the duplicates.  Some of the issues and a respondent to mine mention running the AppImage with "--no-sandbox", but that doesn't work for the deb, which installs "fine" but fails silently from a normal user's perspective. It may be nvidia gpu related, there may be other workarounds ... I use `ddrescue` btw :shrug: 

https://github.com/balena-io/etcher/issues/  

and look for 3873 , which also links to the catalogue of  "duplicates"  :cricket:

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -8,7 +8,7 @@ atom
 audio-recorder
 azure-cli
 azuredatastudio
-balena-etcher-electron
+#balena-etcher-electron
 bat
 battery-monitor
 beersmith3


philclifford:
Also obviously the same deb issue arises through software-boutique @flexiondotorg

natanjunges:
I think the best approach would be to wait. If the problem is solved before the next release of `deb-get`, nothing will need to be done.

philclifford:
> I think the best approach would be to wait. If the problem is solved before the next release of `deb-get`, nothing will need to be done.

It was raised with them almost a year ago. A problem with proton and nvidia gpus apparently. Anyway, I'll close this but  it will be there to help point at the workaround if needed

PR #694, opened 2022-12-11
change mergerfs to show jammy release to kinetic

philclifford:
Because there is no kinetic release, kinetic users get problems for update/show/install.  With this in place they have a better experience.

This uses the available `jammy` release on `kinetic` so we don't break : it works for me, but ymmv - and you may prefer not to do this, or to tweak the `grep` to guess and try and catch and prefer `kinetic` should it appear ...
As it stands it will refuse to update/install as kinetic is not in the supported list.
I've built a kinetic deb to compare and maybe put in a PPA ( upstream only builds and supports releases for LTSs) 
That's probably safer than just using the jammy release - better to use an older version from the ubuntu apt repo than that IMO when it comes to file systems.

### DIFF
--- a/01-main/packages/mergerfs
+++ b/01-main/packages/mergerfs
@@ -1,9 +1,17 @@
 DEFVER=1
 ARCHS_SUPPORTED="amd64 arm64 armhf"
 CODENAMES_SUPPORTED="buster bullseye focal jammy"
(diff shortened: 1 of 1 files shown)


philclifford:
17 hours later the LTS assets appeared so latest works again.

flexiondotorg (approved):
Thanks!

PR #695, opened 2022-12-12
rapid update from github extrepos

philclifford:
~~One API call per repo but noticeably quicker.
Pre-cached json and html files kept alongside extrepo packages will be moved to the cache and touched so they'll be used for a window before being refreshed using API calls~~

Backgrounding  cache maintenance of the supported apps achieves better perceived performance (once it has completed its work - the impatient use of  `list` without `--include-unsupported`  will chug but by the time it has finished the second run will be better)

There is  a useful tweak in https://github.com/wimpysworld/deb-get/pull/695/commits/91bbdaa970cb190070b356e8c3f408810e6b98dd which is possibly worth a cherry-pick.  Allows setting of DEBGET_CACHE_RTN to override the default cache retention time so you can save-for-later or pound-on-the-internet as desired.


philclifford:
Slight bug seems to have sprayed 01-main into my other extra repos .... "one moment please"

PR #718, opened 2023-01-10
fixed bottom

### LINKED ISSUE #716: bottom changed their file naming
they have replaced `$HOST_ARCH` with `$(uname -m)` in their naming scheme so we need to adapt : PR incoming

philclifford:
they changed filenaming
closes #716


philclifford:
And they changed it back ...

PR #731, opened 2023-01-23
Fix insync key removal using keyID fetch

philclifford:
Requires #741 to be merged but then fetches current key correctly.
Existing users with the old key will need to remove it (or uninstall/re-install may do that for them)

### DIFF
--- a/01-main/packages/insync
+++ b/01-main/packages/insync
@@ -1,8 +1,5 @@
 DEFVER=1
-# key gone from website
-# this key is not enough to install but at least keeps _us_ happy
-# # while they sort themselves out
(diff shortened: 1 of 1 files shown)


philclifford:
Rather than use `apt-key` users could do something like this: 
```
sudo gpg --no-default-keyring --keyring /usr/share/keyrings/insync-archive-keyring.gpg --keyserver keyserver.ubuntu.com --recv A684470CACCAF35C
```

we should maybe add an option to have a KEYID in the  app definition and if so fetch it from a keyserver using that kind of pattern

flexiondotorg (approved):

PR #737, opened 2023-02-05
Speed mitigations with no heavy in-repo caching

philclifford:
squashed out caching overload
should include
 - supported app acceleration
 - tunable retention time via env DEBGET_CACHE_RTN
 - tar-based repo update is more efficient

Also removes testing closed of PRs

Without supplying pre-cached info and given 99 githubs to hit and an unauthenticated API limit of 60-per-hour it would seem to be necessary to take the cache out of update completely, but when I tried this I broke it completely.

An alternative idea might be to limit the main repo to say 50 githubs and have the rest on a branch for users with Tokens set .  Chosing the top-50 or even discussing criteria might be "interesting"

### DIFF
--- a/.github/workflows/tests-01-main.yml
+++ b/.github/workflows/tests-01-main.yml
@@ -2,9 +2,12 @@ name: 01-main Tests
 
 on:
   pull_request:
-    types: [opened, closed, synchronize]
+    types: [opened, synchronize]
     paths:
       - "01-main/packages/*"
+      - "!01-main/packages/*.html"
+      - "!01-main/packages/*.json"
+      - "!01-main/packages/timestamp"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/tests-deb-get.yml
+++ b/.github/workflows/tests-deb-get.yml
@@ -2,7 +2,7 @@ name: deb-get Tests
 
 on:
   pull_request:
-    types: [opened, closed, synchronize]
+    types: [opened, synchronize]
     paths:
       - "deb-get"
   workflow_dispatch:
--- a/deb-get
+++ b/deb-get
@@ -160,7 +160,7 @@ function get_github_releases() {
 
     # Do not process github releases while generating a pretty list or upgrading
     if [ "${ACTION}" == "install" ] || [ "${ACTION}" == "update" ] || [ "${ACTION}" == "fix-installed" ]; then
-        if [ ! -e "${CACHE_FILE}" ] || test "$(find "${CACHE_FILE}" -mmin +60)"; then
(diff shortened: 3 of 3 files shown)


flexiondotorg (approved):

yermulnik:
Apparently since the change in this PR `deb-get` started to output useless/extraneous info to stderr from `curl` («fetching non-github repo») and to stdout from `cd -` command:
```bash
> deb-get update
  [+] Updating /etc/deb-get/01-main
/etc/deb-get/01-main.rep 100%[==================================>]   2,70K  --.-KB/s    in 0s
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  102k  100  102k    0     0   172k      0 --:--:-- --:--:-- --:--:--  172k
/home/giermulnik
[---cut---]
```
Can this output be silenced or enabled on-demand please?

yermulnik:
Adding `-s` to `curl` and replacing `cd -` with `cd ~-` should be sufficient to suppress output. Thanks.

yermulnik:
Also silencing down `wget` in the very same `update_repos()` function like it was before (or make it print to stderr on demand) would be great too so that it doesn't spam when it runs e.g. via `cron`. My use case is as simple as this — I want the list of packages pending updates:
```shell
45 9 * * * root deb-get update | fgrep 'has an update pending.'
```
though I don't want this output to be emailed every `deb-get` cron job run (I already silences down `curl` and `cd -`):
```
/etc/deb-get/01-mai   0%[                    ]       0  --.-KB/s
/etc/deb-get/01-mai 100%[===================>]   2,70K  --.-KB/s    in 0s
```

philclifford:
Thanks for your feedback @yermulnik .   :ukraine: 
 
Your cronjob could be `deb-get update 2>/tmp/dg-$$.log | fgrep 'has an update pending.'` or `... 2>/dev/null ...` 

That said I'll work on a PR to shut `cd - ` up, as that's indeed useless noise.  I'll add something to enable turning on "mostly silent mode" optionally somehow. It might get accepted ... Making it quiet by default is unlikely to be acceptable given the output you find useless was deemed useful enough in the general case to specify it.

You'd be better off creating a new issue from a comment above rather than leaving it here as likely no-one but me will read this now the PR has been merged and I ride motorcycles for fun .... :crossed_fingers: 

:blue_heart: 
:yellow_heart:

yermulnik:
@philclifford Thanks for update (and support 🇺🇦).

Redirecting `deb-get` stderr to `/dev/null` was my first idea, though I don't want to redirect stderr as this would just hide any possible or potential errors from my attention.

> You'd be better off creating a new issue from a comment above

Got you. Thanks. Here it is: https://github.com/wimpysworld/deb-get/issues/758

PR #748, opened 2023-02-10
Add installer dependencies

iainelder:
Ubuntu Desktop may already include package `lsb-release` and `wget`. They are not installed in the official `ubuntu:20.04` container image on DockerHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ release pages.
 Use `deb-get` to install `deb-get`.
 
 ```bash
-sudo apt install curl
+sudo apt install curl lsb-release wget
 curl -sL https://raw.githubusercontent.com/wimpysworld/deb-get/main/deb-get | sudo -E bash -s install deb-get
 ```
 


philclifford:
But the deb-get deb will pull in those dependencies ...

iainelder:
I don't see where it pulls these dependencies in. Instead it looks like the main deb-get script expects them to be installed already.

The top level of the script generates an error if lsb_release isn't available.

https://github.com/wimpysworld/deb-get/blob/dd0dff9bdcfa7ae07a14b1021de01af71fc47998/deb-get#L1228-L1230

download_deb assumes that wget is available.

https://github.com/wimpysworld/deb-get/blob/dd0dff9bdcfa7ae07a14b1021de01af71fc47998/deb-get#L220-L229

Here are commands to reproduce it in the Ubuntu container image.

Install curl, then install deb-get:

```bash
docker run -it ubuntu:20.04 bash -c '
apt -y update
apt install -y sudo
sudo apt install -y curl
sudo curl -sL https://raw.githubusercontent.com/wimpysworld/deb-get/main/deb-get | sudo -E bash -s install deb-get
deb-get version
'
```

Eventually it fails like this:

```text
  [!] ERROR! lsb_release not detected. Quitting.
```

Install curl, lsb-release, and wget, then install deb-get.

```bash
docker run -it ubuntu:20.04 bash -c '
apt -y update
apt install -y sudo
sudo apt install -y curl lsb-release wget
sudo curl -sL https://raw.githubusercontent.com/wimpysworld/deb-get/main/deb-get | sudo -E bash -s install deb-get
deb-get version
'
```

Eventually it outputs the version of the newly-installed deb-get:

```text
0.3.7
```

philclifford:
You're absolutely spot on - thanks. If it got as far as the deb ... but it doesn't.  I guess you are the first to install in a container.

PR #750, opened 2023-02-10
Correct master-pdf-editor URL

### LINKED ISSUE #749: master-pdf-editor website oops breaks us
They only updated the redirect and not the manual html link (so far)
The definition needs to be tweaked.

philclifford:
They failed to update their website consistently so we broke, but this forced us to have a better-targeted more focussed search for the download URL
closes #749


Fuseteam:
should we account for cases like this with an additional check? maybe something like `if [ $(echo "${VERSION_PUBLISHED}" | wc -l) -gt 1 ]; then "multiple versions found fi`

Issue #752, opened 2023-02-10
Missing required information on Raspberry Pi


OMEGARAZER:
### Expected behavior

No errors, not seeing download info for 01-main.repo update
(The download part is also seen on my ubuntu 22.04.01 machine but does not get the error related to agena)

### Actual behavior

```
  [!] ERROR! Missing required information of website package agena:
URL=
VERSION_PUBLISHED=
```

### Steps to reproduce

update to 0.3.7 on Raspberry Pi running Debian 11 (5.15.84-v8+ aarch64)

#### /etc/os-release

```bash
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

#### /etc/lsb-release

```bash
cat: /etc/lsb-release: No such file or directory
```

#### deb-get version

```bash
0.3.7
```

### deb-get update

```bash
  [+] Updating /etc/deb-get/01-main
/etc/deb-get/01-main.repo                            100%[===================================================================================================================>]   2.70K  --.-KB/s    in 0s
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  102k    0  102k    0     0   211k      0 --:--:-- --:--:-- --:--:-- 1148k
/var/cache/deb-get
  [+] Updating cache of supported apps in the background
  [!] ERROR! Missing required information of website package agena:
URL=
VERSION_PUBLISHED=
Hit:1 http://security.debian.org/debian-security bullseye-security InRelease
Hit:2 http://deb.debian.org/debian bullseye InRelease
Hit:3 http://deb.debian.org/debian bullseye-updates InRelease
Hit:4 https://download.docker.com/linux/debian bullseye InRelease
Hit:5 http://archive.raspberrypi.org/debian bullseye InRelease
Hit:6 https://deb.nodesource.com/node_19.x bullseye InRelease
Fetched 4,775 B in 2s (2,471 B/s)
Reading package lists...
```

OMEGARAZER:
[...]

PR #756, opened 2023-02-12
Don't bail on unsupported arch

OMEGARAZER:
Adds check for ARCHS_SUPPORTED before bailing validate_deb when arch is unavailable/not supported.

#752

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -471,7 +471,9 @@ function validate_deb() {
 
         if { { { [ "${METHOD}" == github ] || [ "${METHOD}" == website ]; } && [ -s "${CACHE_FILE}" ]; } || [ "${METHOD}" == direct ]; } &&
            { [ "${ACTION}" != "prettylist" ] && [ "${ACTION}" != "remove" ] && [ "${ACTION}" != "purge" ]; } &&
-           { [ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]; }; then
+           { [ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]; } &&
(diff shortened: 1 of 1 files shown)


deb-get
```diff
@@ -471,7 +471,8 @@ function validate_deb() {
 
         if { { { [ "${METHOD}" == github ] || [ "${METHOD}" == website ]; } && [ -s "${CACHE_FILE}" ]; } || [ "${METHOD}" == direct ]; } &&
            { [ "${ACTION}" != "prettylist" ] && [ "${ACTION}" != "remove" ] && [ "${ACTION}" != "purge" ]; } &&
-           { [ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]; }; then
+           { [ -z "${URL}" ] || [ -z "${VERSION_PUBLISHED}" ]; } &&
+           { [[ " ${ARCHS_SUPPORTED} " =~ " ${HOST_ARCH} " ]]; }; then
```
philclifford, line 475: I think it would be necessary/desirable to add codename support check here too: 
```
 { [[ " ${ARCHS_SUPPORTED} " =~ " ${HOST_ARCH} " ]]; }  && { [ -z "${CODENAMES_SUPPORTED}" ] || [[ " ${CODENAMES_SUPPORTED} " =~ " ${UPSTREAM_CODENAME} " ]]; }; then
```

philclifford (commented):
LGTM
Thanks!

OMEGARAZER:
> LGTM Thanks!

I think this gets superseded by #774 does it not?

philclifford:
I think both are needed. I'll take a look later.

philclifford:
This may not be needed now, but IMHO it is better still to be put in place for any unforseen route to calling validate_deb() against an "unsupported app on this platform", or any future behaviour that gets there.  I've pulled your commit into my PR branch and hopefully the `co-authored-by` will preserve your credit if the PR is squashed.

Issue #758, opened 2023-02-14
Decrease `deb-get update` verbositty


yermulnik:
## Requesting a new feature 🧑‍💻

Apparently since the change in https://github.com/wimpysworld/deb-get/pull/737 PR `deb-get` started to output extraneous info to stderr from `wget` and `curl` («fetching non-github repo») and to stdout from `cd -` command (all within `update_repos()` function) — the output between `[+] Updating /etc/deb-get/01-main` and `[---cut---]` lines (which is where starts usual informational output which are not related to this request):
```bash
> deb-get update
  [+] Updating /etc/deb-get/01-main
/etc/deb-get/01-main.rep 100%[==================================>]   2,70K  --.-KB/s    in 0s
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  102k  100  102k    0     0   172k      0 --:--:-- --:--:-- --:--:--  172k
/home/giermulnik
[---cut---]
```
My use case is as simple as this — I want the list of packages pending updates to be emailed to me:
```bash
45 9 * * * root deb-get update | fgrep 'has an update pending.'
```
though I don't want extra output to be emailed every deb-get cron job run. I also would prefer to not redirect stderr to `/dev/null` or a file as this would hide any potential errors which may occur during cron job run.

Can this output be silenced or enabled on-demand please?

How to fix (or make output optional via e.g. new `deb-get update` cmdline parameter):
- Removing `--show-progress` from `wget` cmdline prevents it from outputting download progress (this option forces the progress bar to be printed when `--quiet/-q` option is set)
- Adding `-Ss` to `curl` cmdline prevents it from outputting download progress (while still showing error messages)
- Replacing `cd -` with `cd ~-` prevents `cd` from outputting extra info about previous directory in stack while chdir'ing into it (UPD: https://github.com/wimpysworld/deb-get/pull/759)

### Expected behavior

- `deb-get update` doesn't produce extra output to sdterr/stdout like wit was before https://github.com/wimpysworld/deb-get/pull/737
- Extra output may be turned on or off via command line option.

### Actual behavior

New extra output between `[+] Updating /etc/deb-get/01-main` and `[---cut---]` lines (which is where starts usual informational output which are not related to this request):
```bash
> deb-get update
  [+] Updating /etc/deb-get/01-main
[...]

PR #759, opened 2023-02-14
Make update_repos cd quietly

philclifford:
This should not have been so noisy.  Partially addresses #758 

Co-authored-by: George L. Yermulnik <yermulnik@users.noreply.github.com>


yermulnik:
@philclifford Thanks!
You may also want to look into replacing `cd ${ETC_DIR}/${REPO}.d`/`cd ~-` with `pushd ${ETC_DIR}/${REPO}.d`/`popd`, which may be sort of more Bash-native way 🤔

PR #780, opened 2023-02-26
tests need to authenticate too

philclifford:
The workflows are generating warnings/empty files for all the github-sourced cache files.


philclifford:
Nope

Issue #787, opened 2023-02-28
add Glow app


RichAyotte:
## Requesting a new app ✨

 - [x] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: Glow
 - Download URL: https://github.com/charmbracelet/glow/releases
 - Website URL: https://github.com/charmbracelet/glow
 - App Summary: Glow is a terminal based markdown reader

PR #788, opened 2023-02-28
feat: add glow app

RichAyotte:
## Summary

Add [Glow](https://github.com/charmbracelet/glow) app

Fixes https://github.com/wimpysworld/deb-get/issues/787

## Notes
I wasn't quite sure how to set the `URL` and if there was an easy way to test the changes.

### DIFF
--- a/01-main/packages/glow
+++ b/01-main/packages/glow
@@ -1,11 +1,8 @@
 DEFVER=1
 ARCHS_SUPPORTED="i386 amd64 arm64 armhf"
-get_github_releases "charmbracelet/glow" "latest"
-if [ "${ACTION}" != prettylist ]; then
-    URL="$(grep "browser_download_url.*${HOST_ARCH}\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d '"' -f4)"
-    VERSION_PUBLISHED="$(echo "${URL}" | cut -d '_' -f2)"
-fi
-EULA="MIT License"
+ASC_KEY_URL="https://repo.charm.sh/apt/gpg.key"
+APT_REPO_URL="https://repo.charm.sh/apt/ * *"
+APT_REPO_OPTIONS="arch=${HOST_ARCH}"
 PRETTY_NAME="Glow"
 WEBSITE="https://github.com/charmbracelet/glow"
 SUMMARY="Glow is a terminal based markdown reader designed from the ground up to bring out the beauty—and power—of the CLI."
--- a/01-main/packages/glow
+++ b/01-main/packages/glow
@@ -8,4 +8,4 @@ fi
 EULA="MIT License"
 PRETTY_NAME="Glow"
 WEBSITE="https://github.com/charmbracelet/glow"
-SUMMARY="Glow is a terminal based markdown reader designed from the ground up to bring out the beauty—and power—of the CLI. Use it to discover markdown files, read documentation directly on the command line and stash markdown files to your own private collection, so you can read them anywhere. Glow will find local markdown files in subdirectories or a local Git repository."
+SUMMARY="Glow is a terminal based markdown reader designed from the ground up to bring out the beauty—and power—of the CLI."
--- a/01-main/packages/glow
+++ b/01-main/packages/glow
@@ -1,6 +1,7 @@
 DEFVER=1
 ARCHS_SUPPORTED="i386 amd64 arm64 armhf"
 ASC_KEY_URL="https://repo.charm.sh/apt/gpg.key"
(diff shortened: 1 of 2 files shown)


philclifford:
Hi @RichAyotte , I just drop the app definition into `/etc/deb-get/99-local.d/` and see what `deb-get show glow` comes up with (and install/uninstall). The [test results](https://github.com/wimpysworld/deb-get/actions/runs/4297582992/jobs/7490733104) are prone to false negatives, but a quick look can sometimes help too. Alas today is a day of false negatives: [the irony](https://github.com/wimpysworld/deb-get/actions/runs/4297582992/jobs/7490733104#step:3:35)

01-main/packages/glow
```diff
@@ -0,0 +1,11 @@
+DEFVER=1
+ARCHS_SUPPORTED="i386 amd64 arm64 armhf"
+get_github_releases "charmbracelet/glow" "latest"
+if [ "${ACTION}" != prettylist ]; then
+    URL="$(grep "browser_download_url.*${HOST_ARCH}\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d '"' -f4)"
+    VERSION_PUBLISHED="$(echo "${URL}" | cut -d '_' -f2)"
+fi
+EULA="MIT License"
```
philclifford, line 8: That was a surprise! Not sure if `deborah` will be broken by that. @ymauray may need to have a heads-up if it stays.

philclifford:
While this works perfectly, they recommend (in the Readme Installation section ) for Debian/Ubuntu to use their apt repo, which seems to stay up-to-date with releases, so I'd suggest to use that option rather than github releases.  Then it's just providing the `APT_REPO_URL` and  `ASC_KEY_URL` (and maybe add `APT_REPO_OPTIONS="arch=${HOST_ARCH}"` ).

philclifford (approved):
LGTM
could consider switching to use their `apt` repo source

RichAyotte:
Thanks for the detailed review @philclifford. I'm going to push the change to use the `apt` repo instead shortly.

01-main/packages/glow
RichAyotte: I've removed the EULA for now.

01-main/packages/glow
```diff
@@ -0,0 +1,11 @@
...
+PRETTY_NAME="Glow"
+WEBSITE="https://github.com/charmbracelet/glow"
+SUMMARY="Glow is a terminal based markdown reader designed from the ground up to bring out the beauty—and power—of the CLI. Use it to discover markdown files, read documentation directly on the command line and stash markdown files to your own private collection, so you can read them anywhere. Glow will find local markdown files in subdirectories or a local Git repository."
```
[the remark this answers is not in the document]
RichAyotte: I've shortened the summary

philclifford:
@RichAyotte Thank you!  Just a thought:  as the apt repo is Charm's and may well host other apps that users may install one way or another, it is likely that this approach will lead to conflicts (typically `apt` complaining about conflicting keys ) which we can probably avoid.  For example a quick look confims they have packages there for : 

```
 charm
 confettysh
 glow
 gum
 melt
 skate
 soft-serve
 vhs
 wishlist
```

so if we want to support more than just glow, and to minimise other potential clashes we should use 
`APT_LIST_NAME=glow'
That should provide a reasonable solution.

philclifford:
Yes, confirmed (including on a machine which already had `glow` and `vhs` installed via `apt` ) that the `APT_LIST_NAME=charm` tweak is what's needed : 

```
deb-get show vhs glow |grep -v -e '\['
VHS
  Package:	vhs
  Repository:	99-local
  Updater:	apt
  Installed:	0.3.0
  Architecture:	i386 amd64 arm64 armhf
  Repository:	https://repo.charm.sh/apt/ * *
  Website:	https://github.com/charmbracelet/vhs
  Summary:	A tool for recording terminal GIFs
Glow
  Package:	glow
  Repository:	99-local
  Updater:	apt
  Installed:	1.5.0
  Architecture:	i386 amd64 arm64 armhf
  Repository:	https://repo.charm.sh/apt/ * *
  Website:	https://github.com/charmbracelet/glow
  Summary:	Glow is a terminal based markdown reader designed from the ground up to bring out the beauty—and power—of the CLI.
```

There's still a different key location between their manual 
instructions and where we put it. ( But that's another discussion )

philclifford (approved):
LGTM
Thanks

philclifford (approved):

Adithyan007hack (approved):

PR #810, opened 2023-04-14
6 New Packages

zero77:
firefox
code-insiders
signal-desktop-beta
filezilla-theme-papirus
libreoffice-style-papirus
papirus-folders

### DIFF
--- a/01-main/packages/code-insiders
+++ b/01-main/packages/code-insiders
@@ -1,6 +1,6 @@
 DEFVER=1
 ASC_KEY_URL="https://packages.microsoft.com/keys/microsoft.asc"
-APT_LIST_NAME="vscode beta"
+APT_LIST_NAME="vscode-beta"
 APT_REPO_URL="https://packages.microsoft.com/repos/vscode stable main"
 APT_REPO_OPTIONS="arch=amd64"
 PRETTY_NAME="Visual Studio Code beta"
--- a/01-main/packages/code-insiders
+++ b/01-main/packages/code-insiders
@@ -1,0 +1,8 @@
+DEFVER=1
+ASC_KEY_URL="https://packages.microsoft.com/keys/microsoft.asc"
+APT_LIST_NAME="vscode beta"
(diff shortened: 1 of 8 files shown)


01-main/packages/code-insiders
```diff
@@ -0,0 +1,8 @@
+DEFVER=1
+ASC_KEY_URL="https://packages.microsoft.com/keys/microsoft.asc"
+APT_LIST_NAME="vscode-beta"
```
philclifford, line 3: This must be `"APT_LIST_NAME=code"` or it will potentially lead to having 2 apt list entries with clashing signing key details, breaking both.
However, please also see  https://github.com/wimpysworld/deb-get/blob/main/01-main/CONTRIBUTING.md regarding nightly and beta releases.

philclifford (changes requested):
Thanks for these contributions. Please could you separate this PR out into one-per-package. Otherwise issues over betas or clashing PPAs or APT repos may prevent all of the packages being accepted.

Also please don't update the README.md, that will be done by maintainers for a release and including changes in PRs will create merge conflicts.
Thanks.

01-main/packages/code-insiders
zero77: Are dots or dashes supported as i could use either to separate it from the stable version.
code-insiders or code.insiders

zero77: I have just read CONTRIBUTING.md, is this still the case if it's clearly labeled as being a beta version?

philclifford: > I have just read CONTRIBUTING.md, is this still the case if it's clearly labeled as being a beta version?

I expect it would be rejected because it is labelled beta.  It _might_ be arguable that `code-insiders` is not `beta` so providing the option as well as the stable would be helpful.  Indeed, since they are in the same repository it would be good to have `code-insiders` identified because otherwise if a user installed `code` with `deb-get` they can `apt install code-insiders` but if they then `deb-get remove code` and the apt repo and key will be removed, leaving them in the lurch. So I'd say the case for an exception is strong.

Issue #816, opened 2023-04-18
Add Pulsar


Daeraxa:
- [x] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: Pulsar
 - Download URL: https://github.com/pulsar-edit/pulsar/releases/download/v1.104.0/Linux.pulsar_1.104.0_amd64.deb
 - Website URL: https://pulsar-edit.dev/
 - App Summary: A Community-led Hyper Hackable Text Editor (a community maintained fork of Atom now it has been sunset)

PR #817, opened 2023-04-18
Add Pulsar

Daeraxa:
Fixes: https://github.com/wimpysworld/deb-get/issues/816

I'm one of members of the Pulsar team (https://pulsar-edit.dev/, https://github.com/pulsar-edit) and would love to get this included as one of our earliest entries to a package manager to make it easier for people to install and update on .deb based systems - it is not currently officially available on any Linux package manager - so no PPAs or anything like that.

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -156,6 +156,7 @@ portmaster
 powershell
 protonmail-bridge
 protonvpn
+pulsar
 pulse-sms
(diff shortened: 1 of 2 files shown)


philclifford (approved):
Great to discover that Atom survives the sunset as Pulsar.
LGTM

PR #831, opened 2023-05-14
Rename balena-etcher

LyzardKing:
The package installed is `balena-etcher`, not `balena-etcher-elenctron`
As is an upgrade returns: 
`dpkg-query: package 'balena-etcher-electron' is not installed and no information is available`


philclifford (changes requested):
Also need to change the name in `01-main/manifest`

LyzardKing:
@philclifford Done. I wasn't sure if that was updated automatically

philclifford (approved):

PR #838, opened 2023-06-14
Fix downloads and website urls for Rstudio and Rstudio-server

collioud:
Hello,

URLs for downloading Rstudio and Rstudio server have changed. Updating both repositories to fix this.
I also updated the main website URL from rstudio to posit (new company name).

Tested on my laptop with Ubuntu 20.04. 

Thanks for maintaining deb-get!
Arnaud


philclifford (approved):
LGTM. Thanks

PR #843, opened 2023-06-15
[files-cli] Add the files-cli deb application

arif-ali:


philclifford (approved):
LGTM

philclifford (commented):
@arif-ali You missed (and I failed to spot) adding files-cli to the manifest file, so it has not shown up as supported.

PR #857, opened 2023-07-27
set default timeout and verbosity level for wget invocations

seefood:


deb-get
```diff
@@ -1260,6 +1261,10 @@ function deb_deb-get() {
     SUMMARY="'apt-get' functionality for .debs published in 3rd party repositories or via direct download package."
 }
 
+#### MAIN ####
+
+export WGET_DEFAULT_VERBOSITY="-q"
```
philclifford, line 1266: maybe `export ${WGET_DEFAULT_VERBOSITY:="-q"} `
so that setting WGET_DEFAULT_VERBOSITY in the environment is passed in and not overridden (and same for timeout) ...

seefood: makes absolute sense, I just thought a commandline option with getopt is the right way to do it.
NM, I'll add that, won't hurt.

philclifford (approved):
LGTM
Thanks

deb-get
```diff
@@ -1260,6 +1261,10 @@ function deb_deb-get() {
     SUMMARY="'apt-get' functionality for .debs published in 3rd party repositories or via direct download package."
 }
 
+#### MAIN ####
+
+export ${WGET_VERBOSITY:="-q"}
```
philclifford, line 1266:
```
# cannot do that as part of export - sorry my bad I think!
export CACHE_DIR="/var/cache/deb-get" WGET_VERBOSITY WGET_TIMEOUT
# and then get default values with something like
KLUDGE=${WGET_VERBOSITY:="-q"}
KLUDGE=${WGET_TIMEOUT:="--timeout 5"}
```

deb-get
philclifford: But I was wrong - export barfs on that and you need to do the defaulting separately (see below) Apologies for breaking your contribution.

philclifford (changes requested):
I broke it - sorry.  Needs a tweak

deb-get
seefood: correct. I should have double checked myself too. fixed now.

seefood:
?

philclifford (approved):

philclifford:
Could you please squash these commits,

Thanks

PR #867, opened 2023-08-17
local-apt-repository support

asmbk:
I've created some changes to support local-apt-repository. `deb` package file is copied after install (and before its removal from cache directory) to local-apt-repository.


flexiondotorg:
Thank you for your contribution :bow:  While this is an interesting feature, it is more an advanced capability beyond the scope of `deb-get` which is intended as a simple way to install software from existing release channels.

PR #871, opened 2023-08-30
suggestion: a pre-commit hook test for deb-get

seefood:
If accepted, I'll also commit a fix of the current files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,31 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+---
+# fail_fast: true
+minimum_pre_commit_version: 1.18.1
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+      - id: trailing-whitespace
+        exclude: ".(md|rst)$"
+      - id: end-of-file-fixer
+      - id: check-merge-conflict
+      - id: mixed-line-ending
+      - id: check-added-large-files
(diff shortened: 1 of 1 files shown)


philclifford:
I like this. It does have implications for contributors' systems and workflows as well as potentially improving quality and consistency so needs a bit of testing and discussion and probably some notes in the contribution guide (which needs updating anyway...).
Even if it is not finally accepted and merged I'm pretty sure I'll be retaining and maybe extending these suggestions locally in this and other projects so thanks again for the incentive to explore another area of `git` and foss magic.

flexiondotorg:
I like this idea. I wonder if a better place for these checks might be GitHub actions.

philclifford:
I think it is preferable to have these checks done pre-commit as they help keep all forks clean and shiny. I do have concerns though that imposing `shellcheck` on contributors` systems and workflow might be something better handled in a repository workflow. But then even that simply imposes the installation and use of `shellcheck` on contributors without facilitating that for them, so I end up accepting pre-commit as the place to do this.  

The key for me is accompanying guidance for contributors.

seefood:
Fair enough. I'm all for checks at the IDE level, pre-commit level and the github actions, they each have their function in the workflow. 
As for documentation updates, I'll try to get to all that. I'm just a little pre-occupied here with the war efforts in Israel. May take some time.

philclifford (approved):

flexiondotorg:
Thank you for taking the time to put this together. After careful consideration, we've decided not to move forward with this approach at this time. Closing this PR, but we appreciate your contribution and interest in improving deb-get.

PR #876, opened 2023-09-05
Popcorn desktop

zero77:
No open issues


zero77:
The install fail might be related to: 
https://github.com/popcorn-official/popcorn-desktop/issues/2768

philclifford (changes requested):
The package definition file and the manifest name must match each other, and be identical to the package which the deb installs ( 
 popcorn-time ? ).
Please align those (and also squash the commits and tidy comments) .

PR #877, opened 2023-09-06
Caddy

zero77:
No open issues


felicianotech:
Might be a good idea. I think if this gets accepted, it would be smart to document somewhere that this would be just the core server. The distinction can be important as the Caddy website lets you download custom binaries with your own list of plugins.

philclifford (approved):

philclifford:
There is plenty of documentation and guidance on the linked website and repo.

PR #856, opened 2023-07-23
Simplify+optimize

mralusw:
Hi, I've found `deb-get` really useful and played with it for a bit.

I've tried to optimize performance (by avoiding repeated external calls and caching some results &mdash; e.g. `list` is about 4 times faster), fix some (minor) bugs and simplify some shell constructs (e.g. replacing `grep | cut` with `sed ...`), fix some `shellcheck` warnings etc.

I've tried to make the diffs between commits as small / clear as possible (as opposed to a big rewrite), hence the number of commits.

I've tested it as much as I could on my normal workflows, hopefully github-actions can also help here (I'm not sure how to run testing on my fork).

PR #886, opened 2023-09-14
Adding suggestion: a pre-commit hook test for deb-get

seefood:
bringing the code to compliance following #871


philclifford:
Thanks, that's useful and informative.  Before spending effort resolving these new conflicts maybe wait a while in case #871 receives change requests and/or #856 progresses.  Some of these fixes are also present there and testing both #871 and #856 requires some extra care and thought.  

Also I see the automated testing is having a think: I suspect at least that will be trivial.

flexiondotorg:
Thank you for your work on this. After careful consideration, we've decided not to move forward with this approach at this time. Closing this PR, but we appreciate your contribution and interest in improving deb-get.

PR #893, opened 2023-09-22
Firefox

zero77:
No open issues


01-main/packages/firefox
```diff
@@ -0,0 +1,6 @@
+DEFVER=1
+CODENAMES_SUPPORTED="Mantic Lunar Jammy Focal Bionic Xenial"
```
philclifford, line 2:
```suggestion
CODENAMES_SUPPORTED="mantic lunar jammy focal bionic xenial"
```

philclifford (changes requested):
these need to be lower case

philclifford (approved):

PR #894, opened 2023-09-22
Papirus folders

zero77:
No open issues


01-main/packages/papirus-folders
```diff
@@ -0,0 +1,6 @@
+DEFVER=1
+CODENAMES_SUPPORTED="Kinetic Jammy Focal Bionic Xenial Impish"
```
philclifford, line 2:
```suggestion
CODENAMES_SUPPORTED="mantic lunar kinetic jammy focal bionic xenial impish"
```

philclifford:
It would be good to have a single commit so the addition is atomic, and for the comment to be amended to just refer to the final change. I've also added a suggestion to include mantic, as the ppa supports that.

philclifford (changes requested):

philclifford (approved):

philclifford (approved):

PR #906, opened 2023-10-10
Freetube

zero77:
awaiting a stable release before it can be merged

### DIFF
--- a/01-main/packages/freetube
+++ b/01-main/packages/freetube
@@ -1,0 +1,9 @@
+DEFVER=1
+get_github_releases "FreeTubeApp/FreeTube" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL="$(grep "browser_download_url.*\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d'"' -f4)"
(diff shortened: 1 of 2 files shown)


flexiondotorg (changes requested):
Please rebase.

01-main/packages/freetube
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
```
philclifford, line 1: They are producing amd64, arm64 and armv7l debs
It would be good to add an appropriate ARCHS_SUPPORTED line and incorporate HOST_ARCH in the URL to make them available.

philclifford (changes requested):

01-main/packages/freetube
```diff
@@ -0,0 +1,10 @@
...
+ARCHS_SUPPORTED="amd64 arm64 armv7l"
```
philclifford, line 2: 'armv7l' needs to be `armhf` (ie the relevant option from 
``` shell
dpkg-architecture -L |grep '^arm'
armhf
armel
arm64ilp32
armeb
arm
arm64
```
rather than the HOST_CPU


```suggestion
ARCHS_SUPPORTED="amd64 arm64 armhf"
```

philclifford (changes requested):

PR #908, opened 2023-10-10
Opensnitch

### LINKED ISSUE #682: Add OpenSnitch
 - App Name: OpenSnitch
 - Download URL: https://github.com/evilsocket/opensnitch/releases
 - Website URL: https://github.com/evilsocket/opensnitch
 - App Summary: OpenSnitch is a GNU/Linux port of the Little Snitch application firewall

zero77:
Closes #682

### DIFF
--- a/01-main/packages/opensnitch
+++ b/01-main/packages/opensnitch
@@ -1,0 +1,9 @@
+DEFVER=1
+get_github_releases "evilsocket/opensnitch" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL="$(grep "browser_download_url.*\.deb\"" "${CACHE_FILE}" | grep -v 'python3-opensnitch-ui' | head -n1 | cut -d'"' -f4)"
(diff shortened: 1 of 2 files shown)


01-main/packages/opensnitch
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
+get_github_releases "evilsocket/opensnitch" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL="$(grep "browser_download_url.*\.deb\"" "${CACHE_FILE}" | grep -v 'python3-opensnitch-ui' | head -n1 | cut -d'"' -f4)"
```
philclifford, line 4: fails because the api lists the assets of the release and the latest release is only the UI - there's a link in the release page notes to the latest opensnitch deb but presumably if there's a release of opensnitch without the ui at any point there will be the reverse problem.  I think the only possibility would be to not use get_github_releases but to get the latest release as a web page and parse the link there for each component.  Not ideal but ...

philclifford (changes requested):

PR #910, opened 2023-10-11
feat: Add doublecmd-{gtk,qt}

rasa:


01-main/packages/doublecmd-gtk
```diff
@@ -0,0 +1,8 @@
+DEFVER=1
+CODENAMES_SUPPORTED="buster bullseye focal jammy kinetic lunar"
+ASC_KEY_URL="https://download.opensuse.org/repositories/home:Alexx2000/xUbuntu_${UPSTREAM_RELEASE}/Release.key"
+APT_LIST_NAME="doublecmd-gtk"
```
philclifford, line 4: You need to use the same APT_LIST_NAME or there will be key conflicts (I know most users will be one or the other, but the testing workflow fails because of this.

philclifford (changes requested):

01-main/packages/doublecmd-gtk
rasa: I *think* I fixed what you are referring to. Please let me know if I misinterpreted.

philclifford (approved):

PR #917, opened 2023-10-24
fix: correct curl options when --quiet

### LINKED ISSUE #900: deb-get fails when user's `.curlrc` contains `remote-header-name` option
### Expected behavior

`deb-get update` works

### Actual behavior

`deb-get update` failed. After I removed the `2>/dev/null` that was swallowing the error [here](https://github.com/wimpysworld/deb-get/blob/4303aa855b516079eab9fca467f4eb69a30ac03c/deb-get#L424), curl reports the error
`curl: showing headers and --remote-header-name cannot be combined` 

### Steps to reproduce

```
echo remote-header-name >>~/.curlrc
deb-get update
```

So, I am planning on submitting a PR that adds the `-q` (aka `--disable`) option to all curl calls, so curl doesn't read the user's `.curlrc` file. But I see that the `-q` option has already been added on [line 689](https://github.com/wimpysworld/deb-get/blob/4303aa855b516079eab9fca467f4eb69a30ac03c/deb-get#L689).

So, I'm curious as to why `-q` is added on [line 689](https://github.com/wimpysworld/deb-get/blob/4303aa855b516079eab9fca467f4eb69a30ac03c/deb-get#L689), but not on [line 693](https://github.com/wimpysworld/deb-get/blob/4303aa855b516079eab9fca467f4eb69a30ac03c/deb-get#L693), or any other curl calls.

philclifford:
fixes #900

Co-authored-by: rasa  <rasa@users.noreply.github.com>

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -652,14 +652,12 @@ function refresh_supported_cache_lists() {
 function update_repos() {
     local REPO_URL=""
     # preserve current behaviour for now but allow modification via env
-    local CURL_VERBOSITY="-q --progress-bar"
-    local WGET_VERBOSITY="-q"
+    local CURL_OPTS="--disable --progress-bar"
+    local WGET_VERBOSITY="--quiet"
 
-    if [[ " $* " == *' --quiet '* ]]  ; then
-         CURL_VERBOSITY="-Ss"
-         WGET_VERBOSITY="-q"
+    if [[ " $* " == *' --quiet'* ]]  ; then
+         CURL_OPTS="--disable --show-error --silent"
     fi
-
 
     for REPO in $(find "${ETC_DIR}" -maxdepth 1 -name "*.repo" ! -name 00-builtin.repo ! -name 99-local.repo -type f -printf "%f\n" | sed "s/.repo$//"); do
         # export REPO ETC_DIR ELEVATE  # no longer needed, `| bash -` replaced with `eval`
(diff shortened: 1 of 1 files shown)


deb-get
```diff
@@ -656,7 +656,7 @@ function update_repos() {
     local WGET_VERBOSITY="-q"
 
     if [[ " $* " == *' --quiet '* ]]  ; then
-         CURL_VERBOSITY="-Ss"
+         CURL_VERBOSITY="-Ss -q"
```
rasa, line 659: The `-q` option doesn't quiet curl, `-s` does.

The `-q` option tells curl to not read the user's `.curlrc` file. So, imho, that option should be added to all the curl calls (or even better, passed via an environment variable because some users may want curl to read it's config, so `export CURL_OPTS='-q'`)

philclifford:
Yes, but the -q has been used correctly for the intended purpose to avoid
issues from reading local configs. It probably got lost from the --quiet
case when that was implemented because -s is quiet in curl, but without -q
in both cases there can be issues.

deb-get
philclifford: renamed curl options variable and moved to long naming
also moved --disable to first option because it has to be first

rasa (approved):
LGTM!

flexiondotorg (approved):

PR #918, opened 2023-10-29
Remove Exodus wallet.

marosg42:
They moved download link behind a button press. Original download link has been returning 403 for some time.

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -57,6 +57,7 @@ dustracing2d
 element-desktop
 enpass
 exifcleaner
+#exodus
 expressvpn
 fastfetch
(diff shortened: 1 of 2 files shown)


01-main/manifest
```diff
@@ -56,7 +56,6 @@ duf
 dustracing2d
 element-desktop
 enpass
-exodus
```
philclifford, line 59: Preferable to mark as deprecated ('s/exodus/#exodus/') so users get to know it is no longer being kept updated by `deb-get` and they should consider uninstalling.

philclifford (approved):

PR #935, opened 2023-11-10
manifest: forgot to add jc

mralusw:
Didn't quite realize manifest has to be updated too...

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -107,6 +107,7 @@ irccloud-desktop
 iriunwebcam
 jabref
 jami
+jc
 jellyfin
 jellyfin-media-player
 jexiftoolgui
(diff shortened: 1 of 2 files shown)


philclifford:
No worries. I added it while correcting the docs so this may be superfluous. If I make it home in time I can merge this and revise my PR if needed.

philclifford (approved):

philclifford:
Actually I can't merge it because a test status is required but the test doesn't run on changes only to manifest.  That needs fixing.  Meantime either @flexiondotorg needs to override the requirement or you might be able to provoke a test run by amending the commit to include a change in `packages/jc` (e.g. adding a comment or two from your excellent explanation of the string variable manipulations to help any future contributors improving their `bash` to read )

mralusw:
I've copied the anchoring to the rightmost dquote from #934 (has that PR been missed, BTW?)

PR #971, opened 2024-01-13
Update sunshine codenames to include debian bookworm

hoxu:
Sunshine releases now contain debs for Debian Bookworm:

https://github.com/LizardByte/Sunshine/releases/tag/v0.21.0

I tested this by making the same change to `/etc/deb-get/01-main.d/sunshine` and successfully running `deb-get install sunshine` on Debian Bookworm.

Would be nice to be able to test changes directly from the git checkout.


philclifford (approved):

philclifford:
To test from a checkout just copy the package def from the checkout to /etc/deb-get/99-local.d/ or better still add an external repo definition that points to your fork and branch.

Issue #222, opened 2022-05-30
Add OpenSnitch


dtantono:
I have checked that this app does not already have a pending request and has not previously been declined.
App Name: OpenSnitch
Download URL: https://github.com/evilsocket/opensnitch/releases/download/v1.5.1/opensnitch_1.5.1-1_amd64.deb
Website URL: https://github.com/evilsocket/opensnitch
App Summary: OpenSnitch is a GNU/Linux application firewall.

flexiondotorg:
OpenSnitch is not compatible with 22.04 unless manual workarounds are applied.

- https://github.com/evilsocket/opensnitch/issues/647#issuecomment-1090545284

PR #978, opened 2024-01-23
Python3 opensnitch UI

zero77:
Linked to https://github.com/wimpysworld/deb-get/pull/908

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -182,6 +182,7 @@ peazip
 #picocrypt
 plexmediaserver
 plow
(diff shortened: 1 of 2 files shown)


philclifford (changes requested):
see comments on #908 
This one works for now (although buggy and broken on jammy) but their behaviour suggests it could become the failing half in any release without the ui as an asset if we use the api rather than parse the latest release as a web page.

Also, please squash your commits.

philclifford:
This has previously been declined because of the packaging issues sorry: #222

PR #986, opened 2024-01-29
Popcorn time

zero77:
No open issues

### DIFF
--- a/01-main/packages/popcorn-time
+++ b/01-main/packages/popcorn-time
@@ -1,0 +1,9 @@
+DEFVER=1
+get_github_releases "popcorn-official/popcorn-desktop" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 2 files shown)


01-main/packages/popcorn-time
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
+get_github_releases "popcorn-official/popcorn-desktop" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL="$(grep "browser_download_url.*\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d'"' -f4)"
```
philclifford, line 4: better not to just grab the first deb from a selection - if none specific to OS or release but multiple ARCHs are available then I'd suggest to filter by HOST_ARCH

philclifford, line 4:
```suggestion
    URL="$(grep "browser_download_url.*${HOST_ARCH}\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d'"' -f4)"
```

philclifford (approved):

PR #1025, opened 2024-02-20
Add theme-manager to deb-get

hsbasu:
- Provides theme-manager from launchpad

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -248,6 +248,7 @@ telegraf
 termius-app
 terraform
 texworks
+theme-manager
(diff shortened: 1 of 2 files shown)


flexiondotorg:
Thank you for your contribution, but I am going to politely decline this pull request as desktop environments already provide a means to change the desktop theme.

hsbasu:
I respect your decision. But if I may, this app provides additional functionality like automatically changing the themes at user-defined intervals. Would that be insufficient for this app to be considered?

Issue #1008, opened 2024-02-11
fastfetch has moved to  glibc 2.35 and changed release naming


philclifford:
latest release no longer matches pattern
older releases can be used for older OS releases ( <= focal/bullseye ) until debs build with a compatible glibc are found
The latest release also has an `aarch64` release and adds support for Asahi2, z/OS, Tatra, PikaOS.

mralusw:
There's also a renaming problem, the repo is now `fastfetch-cli/fastfetch`. Though I think LinusDierheimer/fastfetch redirects.

The main problem for me though is that I can't run `deb-get update`, as it chokes with `ERROR! Missing required information of github package fastfetch:` (and also `croc`). If I change `/etc/deb-get/01-main.d/fastfetch`, it reverts back from deb-get repo &mdash; and fails of course. I forgot how to bypass the auto-revert-to-official mechanism

philclifford:
Thanks @mralusw. I didn't notice the redirect so I'll amend the repo in the PR. You can drop the fixed definition into `/etc/deb-get/99-local.d/` (and remember to remove it later when a fix is merged). Or you can add an external repo with a fix. That could be one you maintain (i.e. create a branch on your own fork and add fixes to that branch then add a repo definition to pull that to override 01-main definitions) or you could use my `extras` branch

```
echo "https://raw.githubusercontent.com/philclifford/deb-get/extras/91-fixes" |sudo tee /etc/deb-get/91-fixes.repo
```

You need to remove the "broken" cache files then update, but that should override the broken definitions until main is fixed.  Then you or I need to remove the overrides.

The third option is to submit a PR which I can merge (I cannot merge my own work).

mralusw:
Thanks @philclifford . The override-repo approach seems better. I guess it's documented in the main README. I wonder if I can drop overrides that completely disable annoying packages, by setting some vars to null. I've worked on deb-get before, but I forgot the intricacies of the update process.

philclifford:
[...]

PR #998, opened 2024-02-09
Fix croc

philclifford:
`croc` is not producing deb assets for all releases.

If they don't intend to resume producing debs we will need to deprecate `croc`.
This gets the most recent deb for now and there's nothing obviously significant in subsequent changelogs beyond dependabot, CI tweaks etc.

Closes #997

Issue #997, opened 2024-02-09
"Missing required information of github package croc"


KlfJoat:
Please make sure the version of `deb-get` you are using is up to date:

(excerpt)
```bash
...
Preparing to unpack .../deb-get_0.4.0-1_all.deb ...
Unpacking deb-get (0.4.0-1) over (0.4.0-1) ...
Setting up deb-get (0.4.0-1) ...
...
```
Done.

### Expected behavior

`deb-get update` is resilient against missing information in GitHub packages when the source website changes and the commands return no data or garbage data.

### Actual behavior

`deb-get update` throws an error because yet another package returns no data. This one is `croc`, whatever that is.

```text
[!] ERROR! Missing required information of github package croc:
URL=
VERSION_PUBLISHED=
```

This is a recurring problem (#962  #822) because `deb-get` doesn't gracefully handle the entirely foreseeable and obvious case where the fragile commands used to fill those variables returns nothing or returns garbage. It downloads data from external websites, not versioned APIs. Why is it not resilient against this? Why can't it WARN me and continue processing the rest of the package data? Updates still work, but installs don't. So now there are only 32 packages that users can install from scratch using `deb-get`.

```text
$ deb-get list                                                                      
1password                                      
activitywatch                                                                                       
agena                                                                                               
alduin                                                                                              
android-messages-desktop                                                                            
antimicrox                                                                                          
anydesk                                                                                             
appimagelauncher                                                                                    
atom                                                                                                
atomic                                                                                              
audio-recorder                                                                                      
authme                                                                                              
azuredatastudio                                                                                     
balena-etcher                                             
bat                                                                                                 
bitwarden                                                                                           
bitwig-studio                                                                                       
blanket                                           
blockbench                                                                                          
bluejeans-v2                                                                                        
bottom                                                                                              
brave-browser                                                                                       
brisqi                                                                                              
caddy                                                                                               
caprine
cawbird                                           
chronograf                                        
code                                                                                                
codium
com.github.tkashkin.gamehub                                                                         
copyq                                                                                               
cosign                                            
$
```

### Steps to reproduce

Run `deb-get update`. Look at error on screen.

### System information

Run the following a paste the output below:

#### /etc/os-release

```bash
PRETTY_NAME="Ubuntu 23.10"
NAME="Ubuntu"
VERSION_ID="23.10"
VERSION="23.10 (Mantic Minotaur)"
VERSION_CODENAME=mantic
ID=ubuntu
ID_LIKE=debian
[...]

PR #1027, opened 2024-02-26
fix: skip APP if validate_deb returns false

brccabral:
I have changed `validate_deb()` to return a boolean and if an APP fails update, the rest of the script can continue.  
Apps `croc` #997 #998 and `fastfetch` #1008 are failing right now. Until someone updates their own scripts, `/etc/deb-get/01-main.d/croc` and `/etc/deb-get/01-main.d/fastfetch`, the other apps will continue to update.


philclifford (approved):
Thanks!

Issue #914, opened 2023-10-21
Proposal: Replace non-maintained fork of OBS with the real OBS deb


BrycensRanch:
The package `obs-studio` points to a non-maintained fork of OBS. The successor project, https://github.com/wimpysworld/obs-studio-portable does not yet seem to be available in a deb format yet.

PR #1033, opened 2024-03-03
update obs ppa

brccabral:
I was also going to ask for OBS Studio, should solve #914

### DIFF
--- a/01-main/packages/obs-studio
+++ b/01-main/packages/obs-studio
@@ -1,5 +1,5 @@
 DEFVER=1
-PPA="ppa:flexiondotorg/obs-fully-loaded"
+PPA="ppa:obsproject/obs-studio"
 PRETTY_NAME="OBS Studio (Fully Loaded)"
 WEBSITE="https://obsproject.com/"
 SUMMARY="Free and open source software for video recording and live streaming."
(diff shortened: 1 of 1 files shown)


01-main/packages/obs-studio
```diff
@@ -1,5 +1,5 @@
 DEFVER=1
-PPA="ppa:flexiondotorg/obs-fully-loaded"
+PPA="ppa:obsproject/obs-studio"
 PRETTY_NAME="OBS Studio (Fully Loaded)"
```
philclifford, line 3:
```suggestion
PRETTY_NAME="OBS Studio"
```
This is now the plain upstream without all the extras included.

01-main/packages/obs-studio
```diff
@@ -1,5 +1,5 @@
 DEFVER=1
```
philclifford, line 1:
```suggestion
DEFVER=2
```
Changing the ppa source means this must be updated.

philclifford (changes requested):
Please make the indicated changes (see https://github.com/wimpysworld/deb-get/blob/main/EXTREPO.md re DEFVER)

philclifford (approved):

PR #1042, opened 2024-03-25
option to download oceanaudio using IPv4

brccabral:
`oceanaudio` times out 3 times on IPv6, but IPv4 goes fast

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -178,7 +178,7 @@ function get_website() {
     if [[ ' install update fix-installed ' =~ " ${ACTION} " ]]; then
         if [ ! -e "${CACHE_FILE}" ] || [ -n "$(find "${CACHE_FILE}" -mmin +"${DEBGET_CACHE_RTN:-60}")" ]; then
             fancy_message info "Updating ${CACHE_FILE}"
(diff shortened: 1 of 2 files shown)


deb-get
```diff
@@ -178,7 +178,7 @@ function get_website() {
     if [[ ' install update fix-installed ' =~ " ${ACTION} " ]]; then
         if [ ! -e "${CACHE_FILE}" ] || [ -n "$(find "${CACHE_FILE}" -mmin +"${DEBGET_CACHE_RTN:-60}")" ]; then
             fancy_message info "Updating ${CACHE_FILE}"
-            if ! ${ELEVATE} wget ${WGET_VERBOSITY} --no-use-server-timestamps ${WGET_TIMEOUT} "${1}" -O "${CACHE_FILE}"; then
+            if ! ${ELEVATE} wget ${WGET_VERBOSITY} --no-use-server-timestamps ${WGET_TIMEOUT} "${1}" "${2}" -O "${CACHE_FILE}"; then
```
rasa, line 181: Instead of `"${1}" "${2}"`, `"$@"` would support multiple options, and avoid shellcheck whining about missing options.

flexiondotorg (approved):

PR #1051, opened 2024-04-23
Remove-teams-repo

Fuseteam:
the teams repo is now empty at packages.microsoft.com

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -252,7 +252,7 @@ system-monitoring-center
 tabby-terminal
 tailscale
 tarsnap
-teams
+#teams
 teams-for-linux
 teamviewer
 teip


philclifford (changes requested):
That is not how to remove `teams` from `deb-get`.  Please instead mark the app as deprecated/de-supported by commenting out the entry in 01-main/manifest . ( `sed 's/^teams$/#teams/' 01-main/manifest` )
This will encourage users to de-install the package, which may remove the apt repo from their configs.
Please leave the definition file in place unaltered.  It will be scheduled for removal in due time.

Fuseteam:
understood, will update

philclifford (approved):

PR #1080, opened 2024-05-16
fix: bluejeans has been sunsetted by Verizon

philclifford:
This needs to be deprecated: they've sunsetted it and replaced the website with a message.


philclifford:
External repos /overrides cannot currently deprecate packages.

flexiondotorg (approved):

PR #1083, opened 2024-05-18
Remove old deprecated

philclifford:
Packages removed six months (or a year or more) ago ought to be pruned to avoid annoying deb-get users who may have chosen to install them via deb packages (e.g from the ubuntu repos) with ongoing warnings.

Six months should be adequate time for users to act on the deprecation messages.  Maybe one month would be better.

It might be possible to automate.  My half-cunning process was to identify the candidates from the `git blame` timestamp of the '^#' lines in manifest then

``` bash
for app in  <list_to_be_expunged>
do 
  git rm packages/${app} 
  sed -i "/#${app} /d" manifest
  git add manifest
  git commit -m "chore(${app}): remove packages long deprecated"
done
```

given a way to build find the list it would then be a snip for `gh` to make a branch, do the above loop  and create a PR for review.

Update - CoPilot is quite handy - see #1084 for something that needs polish and refactoring, but might be interesting..


philclifford:
This might also be an indicator that the deprecation warning code could do with a look: I think it should only warn if the app is deprecated in the manifest AND the app is under our aegis, not merely installed from a deb)

flexiondotorg (approved):

PR #1123, opened 2024-07-04
feat: add ticktick

### LINKED ISSUE #1118: Request: TickTick
 - App Name: TickTick
 - Download URL:
   - x64: https://ticktick.com/static/getApp/download?type=linux_deb_x64
   - arm64: https://ticktick.com/static/getApp/download?type=linux_deb_arm64
 - Website URL: https://ticktick.com/
 - App Summary: TickTick is a powerful to-do & task management app with seamless cloud synchronization across all your devices.

philclifford:
closes #1118

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -259,6 +259,7 @@ texworks
 tezos-baking
 thinlinc-client
 thorium-browser
+ticktick
 tidal-hifi
 tixati
 tofu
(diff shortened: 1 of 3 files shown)


philclifford:
Because you need to fetch the url to get cookies but also fetch the filename and version from the head this is a little messy.
There is a function to fetch the filename in the definition which might be able to generalise but probably needs to be specific to the app download provider.

PR #1124, opened 2024-07-06
fix: use unroll_url for ticktick

philclifford:
Remove unneeded app-specific function


philclifford:
@flexiondotorg I bumped the version because I think the changes to correct my bug need to trigger a release.

flexiondotorg (approved):

PR #1137, opened 2024-07-25
fix: fastfetch to use HOST_ARCH

### LINKED ISSUE #1130: bug: fastfetch URL has changed
fastfetch has moved (for some time) from naming release files with HOST_CPU to using HOST_ARCH.
They have just cleaned out the old assets so we now fail to match, although for some time we have been somewhat behind in versions. 

PR Incoming

brccabral:
fastfetch package release naming was changed.
HOST_CPU resolves to `x86_x64`, but the fastfetch releases now have `amd64` which is HOST_ARCH

fixes #1130 
closes #1131


philclifford (approved):

philclifford:
Thanks. This is a duplicate of #1131 and fails to mention that it fixes #1130 (and so is not linked and would not close by reference if merged).

PR #1139, opened 2024-07-26
atuin: No longer releases via debs

rasa:
See https://github.com/atuinsh/atuin/releases and
https://github.com/atuinsh/atuin/pull/1525


philclifford (approved):

philclifford:
already deprecated - this would normally go when deprecated app is cleaned up after a period to allow any users to act on the deprecation warning

Issue #1144, opened 2024-07-31
flameshot fails on Ubuntu 24.04


rasa:
On Ubuntu 24.04 (Noble), flameshot reports
```
  [!] ERROR! Missing required information of github package flameshot:
URL=
VERSION_PUBLISHED=
```
which will be fixed once https://github.com/flameshot-org/flameshot/pull/3678 is merged.

rasa:
https://github.com/flameshot-org/flameshot/pull/3678 has been merged

PR #1147, opened 2024-08-02
fix(flameshot): Empty URL on Ubuntu 24.04 (Noble)

rasa:
Tested and working locally (via 99-local.d)

See #1144


marosg42:
Works for me as well, glad to get rid of of that annoying message.

philclifford (approved):

PR #1148, opened 2024-08-03
feat: move battery-monitor from PPA to github

### LINKED ISSUE #1146: Add Battery Monitor as a .deb package from GitHub instead of PPA
 - App Name: Battery Monitor
 - Download URL: https://github.com/mamolinux/battery-monitor/releases
 - Website URL: https://github.com/mamolinux/battery-monitor
 - App Summary:  An X-platform utility tool developed on Python, notifies about charging, discharging, and critically low battery state of the battery. 

App is already available from deb-get but is sourced from PPA, which makes it impossible to install it on non-Ubuntu distros. Switching to a .deb package would make it available for all Debian based distros, not only *buntus.

philclifford:
closes #1146 

@hsbasu this would seem to be a good option for deb-get users as it widens availability, but I'll leave it draft for now since you moved it to the PPA in #1022 . If you're releasing on github routinely now as well as launchpad I recommend we go ahead but will leave it draft for a short while in case you have concerns.

### DIFF
--- a/01-main/packages/battery-monitor
+++ b/01-main/packages/battery-monitor
@@ -1,7 +1,10 @@
-DEFVER=1
+DEFVER=2
 CODENAMES_SUPPORTED="bookworm bullseye buster trixie sid focal jammy kinetic lunar mantic noble"
-APT_LIST_NAME="apps.mamolinux.org"
-PPA="ppa:mamolinux/gui-apps"
+get_github_releases "mamolinux/battery-monitor" "latest"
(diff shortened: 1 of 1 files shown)


hsbasu:
I do publish deb installers on GitHub for every stable release. So if you think it'd be better to move back to GitHub, you've my green signal.

hsbasu (approved):

philclifford:
Thanks for the quick green light. It is better for those who can't use PPAs.

flexiondotorg (approved):

PR #1169, opened 2024-09-26
fix: get nala key from keyserver

philclifford:
The key has disappeared again.  This fetches it instead by ID from a keyserver.


philclifford:
Given cautions raised by the developer it is better all round I feel that we use #1170

Issue #1098, opened 2024-06-07
 [BUG] nala 'NO_PUBKEY' - fails install


hamensman:
### Expected behavior

Install nala (on fresh Debian 12 install).

### Actual behavior

Fails to install because public key is not available.

### Steps to reproduce

```
$ deb-get install nala     
--2024-06-07 13:30:07--  https://deb.volian.org/volian/scar.key
Resolving deb.volian.org (deb.volian.org)... 99.118.93.168
Connecting to deb.volian.org (deb.volian.org)|99.118.93.168|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2024-06-07 13:30:08 ERROR 404: Not Found.

  [+] Updating: /etc/apt/sources.list.d/nala.list
Get:1 http://deb.volian.org/volian scar InRelease [35.1 kB]
Err:1 http://deb.volian.org/volian scar InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A87015F3DA22D980
Reading package lists... Done
W: GPG error: http://deb.volian.org/volian scar InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A87015F3DA22D980
E: The repository 'http://deb.volian.org/volian scar InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

### System information

Run the following a paste the output below:

#### /etc/os-release

```bash
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

#### deb-get version

```bash
0.4.0
```

### deb-get update

```
  [+] Updating /etc/deb-get/01-main
  [+] Updating cache of supported apps in the background
Hit:1 http://security.debian.org/debian-security bookworm-security InRelease
Hit:2 http://deb.debian.org/debian bookworm InRelease
Hit:3 http://deb.debian.org/debian bookworm-updates InRelease
Hit:4 https://updates.signal.org/desktop/apt xenial InRelease
Hit:5 https://packages.element.io/debian default InRelease
Reading package lists...
```

[...]

PR #1172, opened 2024-09-27
fix: get nala key from gitlab

philclifford:
Retrieve Nala signing key from the GitLab Repo directly.


philclifford:
see #1098

flexiondotorg (approved):

PR #1191, opened 2024-10-18
fix: align beersmith definition to revised website

### LINKED ISSUE #1192: Beersmith have updated their website so the package definition is broken
Legacy versions have also been moved to a separate web page, making fetching and parsing only useful for recent/current releases (which may at some future date become legacy).

philclifford:
They have split legacy versions off to a second page and redesigned the download page so replace dynamic parsing  for now

Fixes #1192


philclifford:
Replaced by single fix in #1208 as test just fails with too many package updates (?)

PR #1211, opened 2024-11-20
fix(nordvpn): reflect changed repository layout

### LINKED ISSUE #1210: NordVPN fails because of repository structure
They have additional layers so we need to dig a little deeper now

philclifford:
fixes #1210


rasa (commented):
The URL is wrong. Use #1213 instead as it fixes the URL too.

PR #1226, opened 2024-12-06
fix(tixati): adapt to changed website structure

### LINKED ISSUE #1225: Tixati fails because of website structure changes
The website has been restructured slightly so we need to adapt

philclifford:
fixes #1225

### DIFF
--- a/01-main/packages/tixati
+++ b/01-main/packages/tixati
@@ -1,8 +1,8 @@
 DEFVER=1
-get_website "https://www.tixati.com/download/linux.html"
+get_website "https://www.tixati.com/linux"
 if [ "${ACTION}" != "prettylist" ]; then
-    URL=$(grep -m 1 "amd64\.deb\"" "${CACHE_FILE}" | cut -d'"' -f2)
-    VERSION_PUBLISHED=$(echo "${URL}" | cut -d'_' -f2)
+    URL=$(grep -m 1 -o  'http.*amd64\.deb\.asc'  /var/cache/deb-get/tixati.html  | sed 's/.asc//')
+    VERSION_PUBLISHED="$(echo "${URL}" | cut -d'_' -f2)"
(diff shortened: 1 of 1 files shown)


flexiondotorg (approved):

01-main/packages/tixati
```diff
@@ -1,7 +1,7 @@
 DEFVER=1
-get_website "https://www.tixati.com/download/linux.html"
+get_website "https://www.tixati.com/linux"
 if [ "${ACTION}" != "prettylist" ]; then
-    URL="$(grep "amd64\.deb\"" "${CACHE_FILE}" | head -n1 | cut -d'"' -f2)"
```
rasa, line 4: @philclifford why move away from `"${CACHE_FILE}"`?

01-main/packages/tixati
```diff
@@ -1,7 +1,7 @@
...
+    URL=$(grep -m 1 -o  '\http.*amd64\.deb.asc'  /var/cache/deb-get/tixati.html  | sed 's/.asc//')
```
rasa, line 4: It pedantic, but I think this wants to be
`URL=$(grep -m 1 -o 'http.*amd64\.deb\.asc'  /var/cache/deb-get/tixati.html  | sed 's/\.asc//')`
to escape the two periods.
The backslash before the h is a typo, yes?

philclifford: yes (it was the one for .asc but this laptop old likes to move cursors around for fun and I wasn't looking where I was typing clearly :-) 
Pedantic is my middle name :-)

01-main/packages/tixati
philclifford: Unintended consequence of tired and hurried attempt to rapidly fix what I broke while also fixing breakages from upstream.   
Sorry.

PR #1246, opened 2024-12-24
add flag to upgrade only deb-get packages

### LINKED ISSUE #1107: [feature request] deb-get upgrade respect --repos-only
## Requesting a new feature 🧑‍💻

If you are proposing a new feature for `deb-get` please describe the proposed
feature with as much detail as possible here:

It would be great if upgrade would respect --repos-only just like update does. When I want to upgrade packages installed through deb-get I neither expect nor want it to update normal packages too.

silentJET85:
closes #1107
This adds the flag `--dg-only` which will only upgrade packages which were installed by deb-get, rather than doing a general apt-get upgrade.
I've tested it in Debian 12 and Kubuntu 24.04 and it seems to work properly.

I used `--dg-only` rather than the `--repos-only` suggested in the feature request, as that flag when updating doesn't trigger apt-get, and I figured that would be inconsistent behavior. Since this *will* trigger apt-get, but only for packages that deb-get has installed. (As listed in /etc/deb-get/installed.)

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ cog.out(f"```\n{help}\n```")
 ]]] -->
 ```
 
-deb-get {update [--repos-only] [--quiet] | upgrade | show <pkg list> | install <pkg list>
+deb-get {update [--repos-only] [--quiet] | upgrade [--dg-only] | show <pkg list> | install <pkg list>
         | reinstall <pkg list> | remove [--remove-repo] <pkg list>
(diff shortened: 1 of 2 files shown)


philclifford (approved):
Not a feature I endorse (the whole point being to apply updates and keep systems up-to-date) but it has been requested and fulfilled so ...

Thanks.

PR #1254, opened 2024-12-29
update package templates

### LINKED ISSUE #1253: Update the package template examples in EXTREPO.md?
## Reporting a bug 🐛
In the package files, it seems like it is now preferred to use `grep -m 1` rather than piping into `head -n 1`, 
and to use "here strings" `<<<` instead of echo, when possible.
It would probably be a good idea to update the package templates to reflect this.

silentJET85:
closes #1253


philclifford (approved):
Very good point!

flexiondotorg (approved):

flexiondotorg (approved):

PR #1278, opened 2025-01-01
fix package versions

### LINKED ISSUE #1277: some packages always report updates available
These packages' version information doesn't match what apt/dpkg says the version number is, so deb-get thinks there are always updates available.

- [x] eversticky
- [x] goreleaser-pro
- [ ] mucommander
- [x] mullvad-vpn

`eversticky` and `mucommander` have a `-1` at the end that shouldn't be there.
`goreleaser-pro`'s URL has the version number with a dash suffix (`-pro`), but apt shows it with a tilde instead.(`~pro`)
`mullvad-vpn` similarly has a tilde instead of a dash for it's beta releases, which we probably shouldn't be grabbing anyways.

silentJET85:
closes #1277

### DIFF
--- a/01-main/packages/eversticky
+++ b/01-main/packages/eversticky
@@ -2,7 +2,7 @@ DEFVER=1
 get_github_releases "itsmejoeeey/eversticky" "latest"
 if [ "${ACTION}" != prettylist ]; then
     URL="$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)"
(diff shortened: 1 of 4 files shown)


philclifford (changes requested):
The CI for `mucommander` only fails because their packaging needs improving. 

Thanks for sorting these. Please could you remove  `mucommander` to a separate PR as this would hopefully then pass.

It would be best to actually make separate PRs for each package, with semantic commits of the form "fix(eversticky): align published version to version in package" This eases maintenance and documentation but most importantly avoids holding up some fixes because of package-specific issues either in the package definition, the packaging or failings in the required tests.

silentJET85:
Done.

PR #1279, opened 2025-01-01
fix(mucommander): align published version to version in package

### LINKED ISSUE #1277: some packages always report updates available

silentJET85:
RE: #1277 #1278
Closes #1277

### DIFF
--- a/01-main/packages/mucommander
+++ b/01-main/packages/mucommander
@@ -2,7 +2,7 @@ DEFVER=1
 get_github_releases "mucommander/mucommander" "latest"
 if [ "${ACTION}" != prettylist ]; then
     URL="$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)"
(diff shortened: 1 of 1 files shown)


philclifford (approved):
This works well but CI fails due to upstream packaging assumptions not applying to the CI environment. (writable xdg system config assumed)

flexiondotorg (approved):

flexiondotorg (approved):

PR #1286, opened 2025-01-02
feat(ci) add version test

### LINKED ISSUE #1285: Feature: ci version test
I think it would be a good idea for the CI test to verify that the installed version number matches the published version number.
This might help prevent packages with unusual version numbers getting stuck in an infinite upgrade loop, as in #1277.

silentJET85:
closes #1285

### DIFF
--- a/.github/workflows/tests-01-main.yml
+++ b/.github/workflows/tests-01-main.yml
@@ -51,7 +51,16 @@ jobs:
                 echo "Failed to detect ${APP} as installed" >&2
                 exit 1
               else
-                ./deb-get purge "${APP}"
+                PUBLISHED_VER=$(./deb-get show "${APP}" | grep "Published:" | grep -oE '[^[:space:]]+$' || :)
+                INSTALLED_VER=$(./deb-get show "${APP}" | grep "Installed:" | grep -oE '[^[:space:]]+$' || :)
+                if [ ! -z "${PUBLISHED_VER// /}" ] && [ "${PUBLISHED_VER}" != "${INSTALLED_VER}" ]; then
+                    echo "Versions do not match. The package's VERSION_PUBLISHED variable may need to be changed." >&2
+                    echo "Published Version: ${PUBLISHED_VER}"
+                    echo "Installed Version: ${INSTALLED_VER}"
+                    exit 1
+                else
+                    ./deb-get purge "${APP}"
+                fi
               fi
             fi
           done
--- a/.github/workflows/tests-01-main.yml
+++ b/.github/workflows/tests-01-main.yml
@@ -52,7 +52,7 @@ jobs:
                 exit 1
               else
                 PUBLISHED_VER=$(./deb-get show "${APP}" | grep "Published:" | grep -oE '[^[:space:]]+$' || :)
-                INSTALLED_VER=$(./deb-get show "${APP}" | grep "Installed:" | grep -oE '[^[:space:]]+$' || :)
+                INSTALLED_VER=$(dpkg-query -Wf '${Version}' "${APP}" || :)
                 if [ ! -z "${PUBLISHED_VER// /}" ] && [ "${PUBLISHED_VER}" != "${INSTALLED_VER}" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-01-main.yml
```diff
@@ -51,7 +51,16 @@ jobs:
                 echo "Failed to detect ${APP} as installed" >&2
                 exit 1
               else
-                ./deb-get purge "${APP}"
+                PUBLISHED_VER=$(./deb-get show "${APP}" | grep "Published:" | grep -oE '[^[:space:]]+$' || :)
+                INSTALLED_VER=$(./deb-get show "${APP}" | grep "Installed:" | grep -oE '[^[:space:]]+$' || :)
```
philclifford, line 55: This might be preferable - just a thought..
```suggestion
                INSTALLED_VER=$(dpkg-query -Wf '${Version}' "${APP}" || :)
```
Marginally more efficient and direct than running `./deb-get` again just to get it to do the above, wrap it in some text and then parse it out again.

.github/workflows/tests-01-main.yml
```diff
@@ -51,7 +51,16 @@ jobs:
                 echo "Failed to detect ${APP} as installed" >&2
                 exit 1
               else
-                ./deb-get purge "${APP}"
+                PUBLISHED_VER=$(./deb-get show "${APP}" | grep "Published:" | grep -oE '[^[:space:]]+$' || :)
+                INSTALLED_VER=$(dpkg-query -Wf '${Version}' "${APP}" || :)
+                if [ ! -z "${PUBLISHED_VER// /}" ] && [ "${PUBLISHED_VER}" != "${INSTALLED_VER}" ]; then
```
philclifford, line 56: I think we should use the same `dpkg` test that deb-get uses for update rather than an exact string comparison here. 
```
if [ "${STATUS}" == "install ok installed" ] && dpkg --compare-versions "${VERSION_PUBLISHED}" gt "${VERSION_INSTALLED}"; then
``` 
so 
```suggestion
                if [ -n "${PUBLISHED_VER// /}" ] && dpkg --compare-versions "${PUBLISHED_VER}" gt "${INSTALLED_VER}" ; then
```

philclifford:
Nice idea, but has the potential to be troublesome so will need a heap of testing (and may generate some work attending to "failures".

.github/workflows/tests-01-main.yml
silentJET85: Wouldn't `ne` be preferable over `gt` in this instance? I'd think it would potentially catch more problems.

philclifford: Actually I think not: it depends on what is a "problem".  I see the "problem" as being that the presented installed version evaluates lower than the presented available version causing persistent "update available". I think it is possible that `ne` would catch false-positives and generate problems where there are none (due to upstreams not following `man deb-versions` or making mistakes).  I think we need to allow for a graceful handling of "close enough" upstream versioning by just using the precise comparison that `deb-get` uses to identify an update.

silentJET85: My thinking was that it could also catch the theoretical opposite scenario, in which the installed version always evaluates higher, and updates never occur. I wonder if something like this might have less false positives than `ne`:
```
if [ -n "${PUBLISHED_VER// /}" ] && dpkg --compare-versions "${PUBLISHED_VER}" gt "${INSTALLED_VER}"  || dpkg --compare-versions "${PUBLISHED_VER}" lt "${INSTALLED_VER}"; then`
```
Or perhaps I'm just overthinking and this unlikely scenario isn't even worth worrying about.

philclifford: I think we're both overthinking a little, but I'd also suggest that the "opposite scenario" is only going to occur if upstream has a problem to fix in their packaging/release processes.

silentJET85: Yeah, that makes sense.

philclifford (approved):

philclifford (approved):

flexiondotorg (approved):

PR #1295, opened 2025-01-14
fix(foliate): change to Github-release method

### LINKED ISSUE #685: Add Foliate as a .deb package from GitHub instead of PPA
 - App Name: Foliate
 - Download URL: https://github.com/johnfactotum/foliate/releases
 - Website URL: https://johnfactotum.github.io/foliate/
 - App Summary:  A simple and modern GTK eBook reader.

App is already available from deb-get but is sourced from PPA, which makes it impossible to install it on non-Ubuntu distros. Switching to a .deb package would make it available for all Debian based distros, not only *buntus.

silentJET85:
closes #685
Closes #1309

This changes from the PPA to the Github-release method. In doing so, support for Trixie/Sid is added, but it loses support for Focal and Jammy.

### DIFF
--- a/01-main/packages/foliate
+++ b/01-main/packages/foliate
@@ -1,5 +1,11 @@
-DEFVER=1
-PPA="ppa:apandada1/foliate"
+DEFVER=2
+CODENAMES_SUPPORTED="trixie sid noble oracular"
+ARCHS_SUPPORTED="amd64 arm64 armhf"
+get_github_releases "johnfactotum/foliate"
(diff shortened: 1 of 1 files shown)


philclifford (dismissed):
It would be better I think to retain the PPA for Ubuntus (or if appropriate, for older ones) and also if running on a Debian or direct derivative make the GitHub source available instead.

silentJET85:
Are we allowed to have multiple methods for one package? I remember that being problematic before. (#521)

philclifford (approved):

PR #1301, opened 2025-01-16
added emptyepsilon

### LINKED ISSUE #1300: App Request: EmptyEpsilon
 - App Name: emptyepsilon
 - Download URL: https://github.com/daid/EmptyEpsilon/releases
 - Website URL: https://daid.github.io/EmptyEpsilon/
 - App Summary: EmptyEpsilon is a spaceship bridge simulator game

gfactor:
Added EmptyEpsilon
Fixes #1300

Please take a close look at how the VERSION_PUBLISHED variable is calculated (line 6 of the emptyepsilon file). I'm unsure if this is the best or most clear way to do it.

### DIFF
--- a/01-main/packages/emptyepsilon
+++ b/01-main/packages/emptyepsilon
@@ -1,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64"
+get_github_releases "daid/EmptyEpsilon" "latest"
+if [ "${ACTION}" != prettylist ]; then
+    URL="$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d'"' -f4)"
+    VERSION_PUBLISHED="$(cut -d'-' -f2 <<< "${URL}" | cut -d'/' -f1)"
+fi
+PRETTY_NAME="Empty Epsilon"
+WEBSITE="https://daid.github.io/EmptyEpsilon/"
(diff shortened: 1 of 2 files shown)


silentJET85:
Looks fine to me. If you're concerned about effeciency, it could also be done with a single sed command. 
For example: `VERSION_PUBLISHED="$(sed -E 's|.*download/EE-([^/]*).*|\1|' <<< "${URL}")"`
The speed difference is negligible, though.

01-main/packages/emptyepsilon
```diff
@@ -0,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64"
+get_github_releases "daid/EmptyEpsilon" "latest"
+if [ "${ACTION}" != prettylist ]; then
+    URL="$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d'"' -f4)"
+    VERSION_PUBLISHED="$(cut -d'-' -f2 <<< "${URL}" | cut -d'/' -f1)"
```
philclifford, line 6: This is OK.  However there may be an issue with how they handle their versioning:
Currently this ends up with
```
  Installed:	2024.12.8
  Published:	2024.12.08
```
which is actually probably OK as .09 will be seen as "higher" than .8, but the inconsistency is a little worrying...

philclifford (approved):

PR #1326, opened 2025-02-17
Add Freetube

zero77:
Replaces the old PR due to problems with rebaseing
https://github.com/wimpysworld/deb-get/pull/906

awaiting a stable release before it can be merged
https://github.com/FreeTubeApp/FreeTube/releases

### DIFF
--- a/01-main/packages/freetube
+++ b/01-main/packages/freetube
@@ -1,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64 arm64 armhf"
+get_github_releases "FreeTubeApp/FreeTube" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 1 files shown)


Xmetalfanx:
ah ... I was going to take a look at this to see why it "failed" but then i saw the message about a "stable release" ... I think a problem (and i do use this app) is that they are always labeled as Beta as Youtube keeps making changes requiring updates for it to work ... this alone maybe why this can't be added.  

Nothing ever crashes for me but sometimes YT makes a change that makes the videos not play, and then the Freetube team has to test/update is all.. but JUST the "beta" label maybe a stumbling block alone for adding this

zero77:
@Xmetalfanx 
I have not tested this, but I think you could still get it to work with beta releases.
But I am not sure if deb-get allows beta releases only stable.

Issue #1336, opened 2025-02-27
ulauncher not working on Debian


soni17:
ulauncher installation not working on Debian because Debian does not support PPA repositories. PPA work only on ubuntu.

Sugguested solution is to direct-download installer from github.

App Name: ulauncher
Download URL: https://github.com/Ulauncher/Ulauncher/releases
Website URL: https://ulauncher.io
App Summary: Application launcher for Linux.

PR #1342, opened 2025-02-28
Add Debian support fo ulauncher

soni17:
Download ulauncher directly from Github. 

Removing PPA repository since it's only supported in Ubuntu and does not work on Debian. 

Fixes https://github.com/wimpysworld/deb-get/issues/1336

### DIFF
--- a/01-main/packages/ulauncher
+++ b/01-main/packages/ulauncher
@@ -1,5 +1,10 @@
 DEFVER=1
-PPA="ppa:agornostal/ulauncher"
+ARCHS_SUPPORTED="amd64 arm64 armhf"
+get_github_releases "Ulauncher/Ulauncher" "latest"
+if [ "${ACTION}" != prettylist ]; then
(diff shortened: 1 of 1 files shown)


01-main/packages/ulauncher
```diff
@@ -1,5 +1,10 @@
 DEFVER=1
```
philclifford, line 1:
```suggestion
DEFVER=2
```

philclifford (changes requested):
Need to update Defver when changing source

01-main/packages/ulauncher
philclifford: EXTREPO.md: `DEFVER`: An integer greater than or equal to `1` that represents the version of the package definition. It must be incremented every time `ASC_KEY_URL`, `GPG_KEY_URL`, `APT_REPO_URL`, `APT_REPO_OPTIONS` or `PPA` are added, changed or removed, or the package installation method changes.

PR #1349, opened 2025-03-05
Update spotify-client pubkey

Phomias:
update spotify pubkey, fetched on Mar 5, 2025 from https://www.spotify.com/us/download/linux/

**Question** 
Do we need to change anything to remove the old pubkey from systems that already had the old one imported?

### DIFF
--- a/01-main/packages/spotify-client
+++ b/01-main/packages/spotify-client
@@ -1,4 +1,4 @@
-DEFVER=2
+DEFVER=3
 ASC_KEY_URL="https://download.spotify.com/debian/pubkey_C85668DF69375001.gpg"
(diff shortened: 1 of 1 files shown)


01-main/packages/spotify-client
```diff
@@ -1,5 +1,5 @@
 DEFVER=2
```
philclifford, line 1:
```suggestion
DEFVER=3
```
We need to update this (see [EXTREPO.md](https://github.com/wimpysworld/deb-get/blob/main/EXTREPO.md) ) so that anyone with a current install via `deb-get` will be prompted to re-install, which should remove the old key and replace it with the new one.

philclifford (changes requested):

philclifford (approved):

PR #1359, opened 2025-03-15
fix: re-fix neo-htop

philclifford:
Reverts wimpysworld/deb-get#1350

Oops : the package name is neo-htop so the manifest was right, the package name needed changing.


rasa (approved):
Sorry, I tested `deb-get update`, I didn't test actually installing it, as I don't use the tool. In the future I will be more thorough on name mismatches.

flexiondotorg (approved):

PR #1362, opened 2025-03-20
fix(jellyfin-media-player): remove arch from jellyfin-media-player

### LINKED ISSUE #1361: jellyfin-media-player has stopped putting HOST_ARCH in the released filename
This breaks our matching.  Simply removing $HOST_ARCH from the match fixes the problem

philclifford:
fixes #1361

### DIFF
--- a/01-main/packages/jellyfin-media-player
+++ b/01-main/packages/jellyfin-media-player
@@ -3,7 +3,7 @@ ARCHS_SUPPORTED="amd64"
 CODENAMES_SUPPORTED="bookworm bullseye kinetic jammy focal noble"
 get_github_releases "jellyfin/jellyfin-media-player" "latest"
 if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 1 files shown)


thestub (approved):

p4p4j0hn (commented):
This looks good. I can confirm it works. I wonder if we should add an `ARCHES_SUPPORTED` line though. It seems like only AMD64 is supported unless something changed.

rasa:
> This looks good. I can confirm it works. I wonder if we should add an `ARCHES_SUPPORTED` line though. It seems like only AMD64 is supported unless something changed.

It's still there:
https://github.com/wimpysworld/deb-get/pull/1362/files#diff-3ec6d301bd3208e6ca447a20635e29e41ce46a6cf96bec8960ccd897117f60bdR2-R3

rasa:
Worked for me too:
```
wget -o /etc/deb-get/99-local.d/jellyfin-media-player https://raw.githubusercontent.com/philclifford/deb-get/refs/heads/fix-jellyfi-media-player/01-main/packages/jellyfin-media-player
deb-get update
deb-get install jellyfin-media-player
```

flexiondotorg (approved):

Issue #1203, opened 2024-11-09
Is the deltachat-desktop website attempt to re-direct broken or does our `unroll_url()` need tweaking?


philclifford:
Deeper question: is their attempt to re-direct broken or does our `unroll_url()` need tweaking?

_Originally posted by @philclifford in https://github.com/wimpysworld/deb-get/issues/1202#issuecomment-2466179561_

PR #1411, opened 2025-04-17
feat(unroll_url): add support for meta refresh

silentJET85:
As mentioned in #1203, `unroll_url` doesn't work on some URLs. One reason for that, is if the site is using the meta refresh tag instead of http redirection. This PR adds support for that.

Here's an example of such a tag:
`<meta http-equiv="refresh" content="1; url=en/download">`


silentJET85:
Done.

PR #1416, opened 2025-04-22
fix(rustdesk): fix ARM support and remove echo

### LINKED ISSUE #1415: [BUG] Rustdesk on arm64
On the latest version of `deb-get` (0.4.5), I'm informed that `rustdesk is not supported on arm64`, but it [is](https://github.com/rustdesk/rustdesk/releases).

Image

Image

---

**EDIT**

After further investigation, it looks like the problem comes from [this line](https://github.com/wimpysworld/deb-get/blob/c8ebc28886aa2a1189f0e6b63faebb63a570c2cf/deb-get#L1330C46-L1330C71). For some reason, when I run `dpkg --print-architecture` on my Raspberry Pi (running Ubuntu 24.04), it outputs `arm64`, which isn't specified in the [ARCHS_SUPPORTED](https://github.com/wimpysworld/deb-get/blob/c8ebc28886aa2a1189f0e6b63faebb63a570c2cf/01-main/packages/rustdesk#L2) variable.

silentJET85:
closes #1415

### DIFF
--- a/01-main/packages/rustdesk
+++ b/01-main/packages/rustdesk
@@ -1,14 +1,9 @@
 DEFVER=1
-ARCHS_SUPPORTED="amd64 armv7 aarch64"
+ARCHS_SUPPORTED="amd64 armhf arm64"
 get_github_releases "rustdesk/rustdesk" "latest"
 if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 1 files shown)


silentJET85:
Hold on, I think I made an arror.

PR #1417, opened 2025-04-22
Fix incorrect supported architectures for RustDesk

StrangeRanger:
**_NOTES:_**

- On Debian based distributions, `dpkg --print-architecture` will never:
    - output `aarch64`, but instead will use `arm64`.
    - output `armv7`, but instead will use `armhf`
    - Screenshot 2025-04-22 at 2 55 58 PM

---

**_PR NOTES:_**

- For the past year, `armv7` releases for `deb` packages have only ever included `armv7-sciter.deb`. As it stood, the script would never be able to get and install the `deb` for the armv7 architecture.
- As mentioned in the above notes, `dpkg` will never output `aarch64` or `armv7`. As a result, the program would never have been able to install `arm` based systems.

---

**_FINAL NOTES:_**

- I have not been able to test the changes directly, so please review them closely to ensure the `grep` patterns is properly implemented. 
- This PR fixes [issue 1415](https://github.com/wimpysworld/deb-get/issues/1415)

### DIFF
--- a/01-main/packages/rustdesk
+++ b/01-main/packages/rustdesk
@@ -1,12 +1,14 @@
 DEFVER=1
-ARCHS_SUPPORTED="amd64 armv7 aarch64"
+ARCHS_SUPPORTED="amd64 armhf arm64"
 get_github_releases "rustdesk/rustdesk" "latest"
 if [ "${ACTION}" != "prettylist" ]; then
     case "${HOST_ARCH}" in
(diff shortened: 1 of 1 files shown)


StrangeRanger:
I have no idea why the test is failing. Maybe because it's trying to save the deb to the `deb-get` cache?

StrangeRanger:
https://github.com/wimpysworld/deb-get/pull/1416 might be prefered over this one, since it passes the checks.

PR #1418, opened 2025-04-23
feat: detect Debian Testing as actual codename instead of Sid

silentJET85:
If someone is running Debian Testing, this should properly detect it by it's codename instead of treating it the same as Sid. 
This should be helpful in cases where a project has separate .debs for Testing and Sid.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -1371,11 +1371,18 @@ if ! [[ ' ubuntu debian ' =~ " ${UPSTREAM_ID} " ]]; then
     fi
 fi
 
-# Debian Sid doesn't have it's own unique os-release file. It is the same as Debian stable.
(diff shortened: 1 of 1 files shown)


silentJET85:
Added a `$DEBIAN_TESTING` variable. If the user is running Debian Testing, this will be the same as `$UPSTREAM_CODENAME`. Otherwise, it will be empty.
 I think this would be useful in instances where an app has a specific build for Debian Testing. OBS Repositories often do, for example. 
This way we wouldn't need to manually adjust the package whenever a new Debian stable comes out and `testing` changes to the next codename.
See #1436 for an example of how it could be used in a package.

I kept it as a separate commit, in case this isn't wanted for whatever reason.

Issue #964, opened 2023-12-14
Enhancement : Make update errors less prominent for apps that are not installed


philclifford:
As noted by @kupiqu in #962 overly noisy reports of errors to users regarding apps that they have not installed can be annoying or confusing.

A possible solution would be to add some sort of "verbose" flag and silence these errors except where the package is installed OR the update is requested to be "verbose".

_Originally posted by @philclifford in https://github.com/wimpysworld/deb-get/issues/962#issuecomment-1856481597_

OMEGARAZER:
> A possible solution would be to add some sort of "verbose" flag and silence these errors except where the package is installed OR the update is requested to be "verbose".

I think this would probably be the best way to go with this as it would hopefully be able to be done in a way that would also solve #757 by only needing to check for things that are requested by/requested by the user.

Issue #757, opened 2023-02-13
0.3.7 Calling github API excessively for each cache update.


OMEGARAZER:
### Expected behavior

Reduced API calls

### Actual behavior

On an update with 0.3.7 the `refresh_supported_cache_lists` function removes the `supported.list` file from the cache which causes a full refresh of information through the lowest branch of the `list_debs` function calling `validate_deb` on every app that is listed as available through installed repos like 01-main with no regard to arch support or installed status. Because of this any incompatible apps were causing #752 (which can be mitigated by #756) and when not hit with #752 will cause 103 calls to the Github API which is well over the 60 available without providing a DEBGET_TOKEN.

### Steps to reproduce

let cache files age over 60 minutes or `deb-get clean && deb-get update` without a DEBGET_TOKEN set and without issue in #752.

#### /etc/os-release

```bash
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

#### /etc/lsb-release

```bash
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"
```

#### deb-get version

```bash
0.3.7
```

### deb-get update

```bash
  [+] Updating /etc/deb-get/01-main
  [+] Updating /var/cache/deb-get/bat.json
  [+] Updating cache of supported apps in the background
  [+] Updating /var/cache/deb-get/deb-get.json
  [+] Updating /var/cache/deb-get/du-dust.json
  [+] Updating /var/cache/deb-get/duf.json
  [+] Updating /var/cache/deb-get/fd.json
  [+] Updating /var/cache/deb-get/gh.json
  [+] Updating /var/cache/deb-get/github-desktop.json
```

### API x-ratelimit-used

Before update: `x-ratelimit-used: 1` (used by the check)
[...]

PR #1419, opened 2025-04-24
feat: only update cache-file for installed apps

silentJET85:
This PR will make it so that  `deb-get update` will only update the cache file for apps that are already installed.
Unless I'm overlooking something, there doesn't seem to be any need to have the cache file for non-installed apps. When installing an app, it will still be downloaded if needed.

This should significantly reduce the amount that the Github API is being called for the average user, at least partially solving #757. As well as just speeding up the `deb-get update` process overall.

This should also have the side-benefit of reducing the errors shown for apps that are not installed, as mentioned in #964.

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -164,7 +164,7 @@ function get_github_releases() {
     #   curl -I https://api.github.com/users/flexiondotorg
 
     # Do not process github releases while generating a pretty list or upgrading
-    if [[ ' install update fix-installed ' =~ " ${ACTION} " ]]; then
+    if [[ ' install fix-installed ' =~ " ${ACTION} " ]] || { [[ ' update ' =~ " ${ACTION} " ]] && package_is_installed "${APP}"; }; then
         if [ ! -e "${CACHE_FILE}" ] || [ -n "$(find "${CACHE_FILE}" -mmin +"${DEBGET_CACHE_RTN:-60}")" ]; then
             fancy_message info "Updating ${CACHE_FILE}"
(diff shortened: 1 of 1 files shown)


GideonBear:
Applied the patch on my system, no problems so far. Good idea!
One thing: it does hide errors for apps that are not installed, which could also hide the errors from maintainers; it's probably best to catch those errors early. Maybe we could set up a workflow to run daily that updates the cache for all packages (or maybe even installs all packages)?

silentJET85:
I've been tossing around the idea of making a local script to do something similar in virtual machines on a spare computer. Maybe even having a separate VM for each supported release (bookworm, trixie, jammy, noble, etc) and testing every package in each, to be thorough.

But the Github Workflow idea is probably better and less maintenance. My only worry would be about exceeding the allotted minutes if we install every package. (Not sure what tier they're on.) Although installing them all would be helpful in finding breakages early.

Perhaps just updating the cache daily, and doing full installs only weekly or monthly would be sufficient. Or something like that.

GideonBear:
Agreed that it *should* do it on all supported releases, but unfortunately it doesn't look like GitHub Actions has them all, and I don't think it's a good idea to run it in containers, since packages might depend on things like graphical libraries, etc. to be available.
Also note that installing the same version of a package twice is absolutely pointless (right?), so we should be able to cache the version that was last tested, and only test new versions of packages; we could probably run it daily if we do it that way.

PR #1494, opened 2025-08-14
feat: add linuxtoys

brccabral:
LinuxToys  
A collection of tools for Linux in a user-friendly way.  
https://github.com/psygreg/linuxtoys


01-main/packages/linuxtoys
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
+get_github_releases " psygreg/linuxtoys" "latest"
```
philclifford:
```suggestion
get_github_releases "psygreg/linuxtoys" "latest"
```

PR #1499, opened 2025-08-21
Add env variable to disable apt

tmiland:
- Added env variable to be able to disable apt

To be used with [https://github.com/tmiland/deb-get-apt-hook](https://github.com/tmiland/deb-get-apt-hook)


flexiondotorg:
@cubic-dev-ai Review this

PR #1504, opened 2025-08-29
fix(picocrypt-gui):  depends on later libc6 version than available in Bookworm/Jammy

philclifford:
Not reflected in deb control so installs but fails to run.
Needs libc6 >=2.38


philclifford:
Obsoleted  by #1539

PR #1531, opened 2025-09-28
added Webmin

brccabral:
Webmin is a web-based system administration tool for Unix-like servers, and services.  
https://webmin.com


philclifford (commented):
It would be better to add the apt repository I think, since the sourceforge downloads are described as "older releases" in the installation section of the Readme on GitHub.

brccabral:
Yes, I've found this tool through an old blog.

Issue #1550, opened 2025-10-21
App Request: nicebucket


moschlar:
### Is this a new request?

- [x] I have checked that this app does not already have a pending request and has not previously been declined.

### App Name

nicebucket

### Download URL

https://github.com/nicebucket-org/nicebucket/releases

### Website URL

https://github.com/nicebucket-org/nicebucket

### App Summary

 A fast, private, open-source S3 GUI.

philclifford:
Thanks. But I don't think 0.0.10 is considered a stable/production release so I'll mark the PR as draft for now. When they're further down the roadmap it can be reconsidered
See https://github.com/wimpysworld/deb-get/blob/main/01-main/CONTRIBUTING.md

PR #1551, opened 2025-10-21
Add nicebucket

moschlar:
Closes: #1550


philclifford (changes requested):
You have to also add an entry in the manifest

philclifford:
Still an early development version.

PR #1567, opened 2025-11-06
 Flameshot add support for debian sid

zero77:
add support for debian sid

### DIFF
--- a/01-main/packages/flameshot
+++ b/01-main/packages/flameshot
@@ -1,5 +1,5 @@
 DEFVER=1
-CODENAMES_SUPPORTED="buster bullseye trixie focal jammy mantic noble oracular"
+CODENAMES_SUPPORTED="buster bullseye trixie focal sid jammy mantic noble oracular"
 get_github_releases "flameshot-org/flameshot" "latest"
 if [ "${ACTION}" != "prettylist" ]; then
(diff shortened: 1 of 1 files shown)


01-main/packages/flameshot
```diff
@@ -1,5 +1,5 @@
 DEFVER=1
-CODENAMES_SUPPORTED="buster bullseye trixie focal jammy mantic noble oracular"
+CODENAMES_SUPPORTED="buster bullseye trixie focal sid jammy mantic noble oracular"
```
philclifford, line 2:
```suggestion
CODENAMES_SUPPORTED="bookworm jammy noble"
```
They only have [release assets](https://github.com/flameshot-org/flameshot/releases/latest) for Debian 12 and Ubuntu 22.04 and 24.04.

zero77:
I have been using it on a Debian sid device for a while and not had any problems.

```
cat /etc/os-release

PRETTY_NAME="Debian GNU/Linux forky/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=forky
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

philclifford:
Yes, but the definition is written to match the available debian and ubuntu release assets so putting other releases in CODENAMES_SUPPORTED will just lead to 404 errors and `deb-get` will never succeed. There is nothing to stop trixie , sid or questing users from adding their own override to map an available release asset but that's not an approach we should take in main.

zero77:
ok i will stick to changing it manually with something like sed

Issue #1554, opened 2025-10-23
chore: review explicit release support now that questing(25.10) has been released


philclifford:
> Need to 
> 
> - [ ] review/update definitions where explicit release support restrictions exist 

 _Originally posted by @philclifford in [#1386](https://github.com/wimpysworld/deb-get/issues/1386#issuecomment-3438402740)_

- [x] Remove unsupported interim releases [^1]
- [x] Find and update any where support is in place
- [x] Prepare list of pending updates

----------

<details>
- [ ] flowkeeper
- [ ] fsearch
- [x] input-leap (awaiting upstream release fix)
- [ ] jellyfin-media-player
- [ ] libation
- [x] librewolf
- [ ] master-pdf-editor-5
- [ ] media-downloader
- [ ] mergerfs
- [ ] museeks
- [ ] nala
- [ ] oculante
- [ ] openrgb
- [ ] papirus-folders
- [ ] pet
- [ ] proton-pass
- [ ] qownnotes
- [ ] simple-pwgen
- [ ] sunshine
- [ ] terratactician-expandoria
- [ ] vagrant
- [x] veracrypt
- [x] veracrypt-console
- [ ] waydroid
- [ ] winegui
- [ ] winehq-devel
- [ ] winehq-stable
- [ ] winehq-staging

</details>
These need monitoring

- [ ] flameshot
- [ ] i3 (plucky support there - questing not yet)
- [ ] input-leap (awaiting upstream release workflow fix)

These are done/pending PR/nfa

 - [x] alsa-scarlett-gui
 - [x] azure-cli
 - [x] battery-monitor
 - [x] beersmith3
 - [x] cudatext (needs PR or newer versions get old QT )
 - [x] flameshot (fix : only debian 12 , jammy and noble assets currently)
 - [x] firefox*
 - [x] rstudio*  (only /jammy/ files so restriced to jammy+/bookworm+
 - [x] Tonelib-*
 - [x] veracrypt
 - [x] veracrypt-console
 - [x] virtualbox-6.1 (ignore - pending deprecation)
 - [x] virtualbox-7.0 (ignore - pending deprecation)

( see PRs tagged below )
----
 
These may need to be deprecated

 - [ ] quickobs
 - [ ] deborah
 - [ ]

philclifford:
Odds are a single bundled PR updating all/most of these definitions with fail CI currently. 
Scripting one PR per app is probably do-able but is it worth it ?

PR #1576, opened 2025-11-07
chore: update supported releases - winehq

philclifford:
working on #1554


philclifford:
Fails to resolve dependencies for -devel.  I did an upgrade the other day from jammy to noble and winehq-stable failed to re-install/upgrade because it seems to have developed a dependency on wine that is not resolved until you install wine (from the ubuntu noble repo) ... pretty sure no such fun when I updated some systems to noble a year or so back or with any of their upgrades ....

Will do some more headbanging around this but setting draft until fixed and/or split up

PR #1590, opened 2025-11-07
chore: update supported releases - foliate

philclifford:
working on #1554


philclifford:
foliate deb fails to install on noble (foliate : Depends: gir1.2-adw-1 (>= 1.7) but 1.5.0-1ubuntu2 is to be installed) 
foliate is now in debian/ubuntu repos and the author recommends using his snap (or flatpak)

We should therefore de-support the package

Issue #1556, opened 2025-10-26
Repository change for Haguichi


ztefn:
@silentJET85 @flexiondotorg I have created a new repository for Debian packages using openSUSE Build Service:
https://haguichi.net/download/#debian
https://build.opensuse.org/package/show/home:ztefn:debian/haguichi

The problem with the current [Haguichi Debian PPA](https://launchpad.net/~ztefn/+archive/ubuntu/haguichi-debian) is that there is no Ubuntu LTS that matches Debian 13. Ubuntu 25.04 (plucky) is the closest release, but that one is only available to build packages for on Launchpad for 9 months. Deb-get currently even uses Ubuntu 24.10 (oracular) which is already EOL and cannot receive any new builds on Launchpad. So there is no sane path forward on Launchpad to target Debian 13 and I will remove all 1.5.x packages in a few months when Ubuntu 25.04 is also EOL. Only the legacy 1.4.6 packages for Debian 12 will be kept in there.

For Ubuntu packages the [Haguichi Stable PPA](https://code.launchpad.net/~ztefn/+archive/ubuntu/haguichi-stable) will remain unchanged.

PR #1591, opened 2025-11-07
fix(haguichi): move debians onto plucky in ppa

philclifford:
working on #1554


ztefn:
Please don't. The repository for Debian packages has recently been changed to OBS so that there is no weird Ubuntu code name matching shenanigans needed. See #1556.

philclifford:
Thanks for catching my regression attempt!

PR #1618, opened 2025-11-08
chore: update supported releases - tonelib-bassdrive

philclifford:
working on #1554


philclifford:
All ToneLib packages still have dependencies on packages that have been removed from noble.

Issue #1076, opened 2024-05-12
Ad LACT (Linux AMDGPU Controller)


sbstn87:
- [X] I have checked that this app does not already have a pending request and has not previously been declined.
 - App Name: LACT - Linux AMDGPU Control Application
 - Download URL: https://github.com/ilya-zlobintsev/LACT/releases
 - Website URL: https://github.com/ilya-zlobintsev/LACT
 - App Summary: This application allows you to control your AMD GPU on a Linux system. Current features:
    Viewing information about the GPU
    Power/thermals monitoring
    Fan curve control
    Overclocking (GPU/VRAM clockspeed, voltage)
    Power states configuration

philclifford:
Support for 24.04 is in testing and a new release may be arriving shortly. Possibly worth holding on for that to save having to re-rev the definition.

sbstn87:
A new version supporting RDNA4, Ubuntu 24.04 has been recently released.

PR #1671, opened 2026-02-04
Add package: LACT

nrocha02:
Closes: #1076

### DIFF
--- a/01-main/packages/lact
+++ b/01-main/packages/lact
@@ -1,7 +1,7 @@
 DEFVER=1
 ARCHS_SUPPORTED="amd64"
 CODENAMES_SUPPORTED="bookworm jammy noble"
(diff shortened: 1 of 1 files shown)


01-main/packages/lact
```diff
@@ -0,0 +1,12 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64"
+CODENAMES_SUPPORTED="bookworm jammy noble"
+UPSTREAM_RELEASE_NODOT=$(echo "$UPSTREAM_RELEASE" | tr -d '.')
```
philclifford:
```suggestion
UPSTREAM_RELEASE_NODOT="${UPSTREAM_RELEASE//./}"
```

Bash substitution removes need for echo and tr.

Issue #1678, opened 2026-02-10
App Request: dra


salim-b:
### Is this a new request?

- [x] I have checked that this app does not already have a pending request and has not previously been declined.

### App Name

dra

### Download URL

https://github.com/devmatteini/dra/releases

### Website URL

https://github.com/devmatteini/dra

### App Summary

A command line tool to download release assets from GitHub.

PR #1679, opened 2026-02-11
feat: add dra

salim-b:
closes https://github.com/wimpysworld/deb-get/issues/1678


01-main/packages/dra
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
+get_github_releases "devmatteini/dra" "latest"
```
philclifford, line 4:
```suggestion
# depends on libc6 (>= 2.39)
CODENAMES_SUPPORTED="trixie sid noble questing"
get_github_releases "devmatteini/dra" "latest"
```

Issue #1738, opened 2026-02-19
Bug: github api output has changed


philclifford:
several github sourced packages error on update 
The output json is one line so need splitting.

EDIT: and the api response seems to have reverted and the problem has gone away (for now)

Still may be worth considering procatively reformating the results in case the behaviour re-occurs. This is now less urgent than it was a few hours ago so removing the bug label.

philclifford:
Packages with issues (a few are possibly other errors but ...)
(checked ones I have WIP fixes on the way ...)
  - [x] activitywatch
  - [x] bat
  - [x] bottom
  - [x] deemix
  - [x] dbgate
  - [x] dorion
  - [x] duf
  - [x] ente
  - [x] fastfetch
  - [x] filen
  - [x] flameshot
  - [x] flowkeeper
  - [x] ganttproject
  - [x] gh
  - [x] gk
  - [x] goreleaser
  - [x] grype
  - [x] helmwave
  - [x] hugo
  - [x] ipscan
  - [x] jellyfin-media-player
  - [x] jitsi-meet-electron
  - [x] joplin
  - [x] kse
  - [x] libation
  - [x] manuskript
  - [x] mdview
  - [ ] mqttx
  - [ ] mucommander
  - [ ] museeks
  - [ ] nuclear
  - [ ] obsidian
  - [ ] pandoc
  - [ ] pet
  - [ ] pixieditor
  - [ ] quarto
  - [ ] quickgui
  - [ ] rclone
  - [ ] readest
  - [ ] r-quick-share
  - [ ] rustdesk
  - [ ] slackdump
  - [ ] sleek
  - [ ] spotube
  - [ ] stirling-pdf
  - [ ] syft
  - [ ] teams-for-linux
  - [ ] tinygo
  - [ ] tribler
  - [ ] trilium
  - [ ] ulauncher
  - [ ] wasistlos
  - [ ] xpipe
  - [ ] yazi
  - [ ] zenith


```
  Package:	activitywatch
  Published:	178523506
  Download:	https://api.github.com/repos/ActivityWatch/activitywatch/releases/178523506
  Package:	bottom
  Published:	273658454
  Download:	https://api.github.com/repos/ClementTsang/bottom/releases/273658454
  Package:	deemix
  Published:	https://api.github.com/repos/bambanah/deemix/releases/276198854
  Download:	https://api.github.com/repos/bambanah/deemix/releases/276198854
--
  Package:	dorion
  Published:	274083237
[...]

PR #1740, opened 2026-02-19
fix: convert minified Github JSON back into pretty-print

silentJET85:
RE: #1738
This PR ensures that the Github releases JSON is in the "pretty-print" format, rather than all being on one line. So that existing packages can continue to parse it correctly. 
closes #1739

Only downside is, I don't think `jq` is installed by default for most people. So we would need to add it to the list of dependencies.


philclifford (changes requested):
please add jq to the dependencies in debian/control

philclifford:
Much better than the alternative huge chore I was ploughing through. I think it is worth the small price of and extra dependency.

philclifford (approved):
Thanks!

Issue #1699, opened 2026-02-13
bug(winegui):  now fails CI test due to upstream


philclifford:
### Other Issue

see #1693

silentJET85:
There seems to be two problems:

1. They now have separate .deb files for each version of Debian and Ubuntu, instead of one deb for everything. That's easy enough to fix. (Although they don't have debs for Sid and Questing, the Forky and Plucky debs install fine in them, respectively.)

2. The other problem is that their server seems to be rejecting the connection when using `wget`. The same URL works fine in a browser.
Here's the result when I try to download the assets json manually:
```
wget -v --no-use-server-timestamps "https://gitlab.melroy.org/api/v4/projects/66/releases/v3.0.0/assets/links" -O-                                                                      
--2026-02-14 21:27:23--  https://gitlab.melroy.org/api/v4/projects/66/releases/v3.0.0/assets/links   
Resolving gitlab.melroy.org (gitlab.melroy.org)... 77.61.56.117                                      
Connecting to gitlab.melroy.org (gitlab.melroy.org)|77.61.56.117|:443... failed: Connection refused.
```

EDIT: Their site is also not loading for me in Waterfox and Ungoogled Chromium.  It also won't connect if I switch to a Private Window in Brave. This makes me think it won't be fixable on our end.

philclifford:
```text
$ wget  -v --no-use-server-timestamps  https://winegui.melroy.org/downloads/WineGUI-v3.0.0-plucky.deb
--2026-02-18 11:28:56--  https://winegui.melroy.org/downloads/WineGUI-v3.0.0-plucky.deb
Resolving winegui.melroy.org (winegui.melroy.org)... 77.61.56.117
Connecting to winegui.melroy.org (winegui.melroy.org)|77.61.56.117|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 736030 (719K) [application/octet-stream]
Saving to: ‘WineGUI-v3.0.0-plucky.deb.1’

WineGUI-v3.0.0-plucky.deb.1                              100%[===>] 718.78K  2.59MB/s    in 0.3s    

2026-02-18 11:28:57 (2.59 MB/s) - ‘WineGUI-v3.0.0-plucky.deb.1’ saved [736030/736030]

```

the refusal seems to have been transient 🤞 

Hopefully fixing up the new filenames (and setting the supported codenames) will be sufficient .

philclifford:
No - it's back : 
```text
[...]

PR #1741, opened 2026-02-20
feat: add user-agent to wget calls

### LINKED ISSUE #1737: feat: add a "--user-agent" parameter to `wget` calls
Seems like it might be helpful (and polite) to differentiate our visits from the noise. 

> I think if you would set a custom user-agent string, instead of the default `wget` that would already help to identify your tool better at reverse proxy level / fail2ban etc. By simply using `--user-agent=deb-get/<version>` parameter in this case.

 _Originally posted by @melroy89 in [#1699](https://github.com/wimpysworld/deb-get/issues/1699#issuecomment-3922858206)_

philclifford:
closes #1737 

see #1699

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -186,7 +186,7 @@ function get_github_releases() {
                 fancy_message info "Updating ${CACHE_FILE}"
             fi
             local URL="https://api.github.com/repos/${1}/releases${2:+/$2}"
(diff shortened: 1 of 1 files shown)


philclifford:
Before going ahead with this now the driving issue is resolved I think a bit more thought, discussion and assurance are worthwhile.

PR #1761, opened 2026-03-04
fix: LACT are moving to  debian-13 soon

### LINKED ISSUE #1757: LACT now works with Debian 13
### Other Issue

LACT (including the 'debian-12' versions) works fine with Debian 13.  ..it also appears that they're renaming the package to state 'debian-13' instead of 'debian-12', as per their recent test release.  I don't know if they're retaining Debian 12 compatibility.

philclifford:
We will need to remove support for unsupported releases at some point. When they release a trixie-based version this  update can be released. If it is before the release of Ubuntu 26.04 including resolute may be premature

Closes #1757


eode:
They have a debian-13 file up now, if this is waiting on that.

philclifford:
Borked branch - replaced by #1909

PR #1784, opened 2026-03-11
add package: wifiman-desktop

### LINKED ISSUE #1772: App Request: wifiman-desktop
### Is this a new request?

### App Name

wifiman-desktop

### Download URL

https://desktop.wifiman.com/wifiman-desktop-1.2.8-amd64.deb

### Website URL

https://ui.com/download/app/wifiman-desktop

### App Summary

WiFi analyzer and Teleport (VPN) client by Ubiquiti Networks.

snoack:
Closes #1772

### DIFF
--- a/01-main/manifest
+++ b/01-main/manifest
@@ -396,6 +396,7 @@ weechat
 weylus
 whalebird
 #whatsapp-for-linux
+wifiman-desktop
 winegui
 winehq-devel
 winehq-stable
--- a/01-main/packages/wifiman-desktop
+++ b/01-main/packages/wifiman-desktop
@@ -1,0 +1,20 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64"
+CODENAMES_SUPPORTED="bookworm trixie forky sid jammy noble plucky questing"
+get_website "https://desktop.wifiman.com/wifiman-desktop-linux-manifest.json"
+if [ "${ACTION}" != "prettylist" ]; then
+    VERSION_PUBLISHED="$(grep -ozP '"version"\s*:\s*"\K[^"]+' "${CACHE_FILE}")"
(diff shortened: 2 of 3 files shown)


snoack:
About the CI failure, apparently the `postinst` script runs `ps aux | grep 'wifiman-desktop' | awk '$1!="root"{print $2}' | xargs -r kill -SIGTERM` in an attempt to terminate running instances of the app but instead it kills `deb-get install wifiman-desktop`. The install seems to be successful though.

Not sure what to do about it, maybe merge it as is (and hope that upstream will fix it eventually)? I'd rather have `deb-get install wifiman-desktop` terminate with non-zero exit status after successful install than having to install and update `wifiman-desktop` manually.

A workaround could be renaming the package definition file (e.g. to just `wifiman`) but I'm not sure if there are any side effects of using a different name for the package definition in `deb-get` than the package in `apt`?

philclifford:
The package definition name must exactly match the apt package name. The side effect of a different name would be never seeing the package as installed so never detecting or attempting an update (and failing to uninstall the failed install because the name is not found).
I suggest raising the problem upstream so that their post install behaviour succeeds without errors or handles errors appropriately.

01-main/manifest
```diff
@@ -400,6 +400,7 @@ winegui
 winehq-devel
 winehq-stable
 winehq-staging
+wifiman-desktop
```
philclifford: please move this up - the manifest entries should be sorted alphabetically

snoack:
Alright, as a workaround I have introduced a `post_download` hook in order to rebuild the package with the buggy `postinst` script patched. Note that the `post_download` hook mechanism would not only be useful here but is also required for #1782.

01-main/manifest
snoack: Done.

PR #1796, opened 2026-03-16
Add "forky" to CODENAMES_SUPPORTED for all packages that support "trixie" and "sid"

snoack:
If they support the current stable and unstable they are certainly supporting the next stable which is in between.

### DIFF
--- a/01-main/packages/alt-sendme
+++ b/01-main/packages/alt-sendme
@@ -1,5 +1,5 @@
 DEFVER=1
-CODENAMES_SUPPORTED="bookworm trixie sid jammy noble questing"
+CODENAMES_SUPPORTED="bookworm trixie forky sid jammy noble questing"
 get_github_releases "tonyantony300/alt-sendme" "latest"
 if [ "${ACTION}" != prettylist ]; then
     URL="$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)"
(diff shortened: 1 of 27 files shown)


snoack:
If you want I can remove the packages that are failing from the scope of this change but I'm rather confident the failures are unrelated of this change.

philclifford:
Failures are unrelated and being looked at. Please leave them in if appropriate.

philclifford:
Your assumption of 'certainly support next stable' is tempting and likely true of Debian and Ubuntu packaging but 3rd party repos and release assets will invariably find a way from time to time to achieve breakages that cannot happen. Since we only deal with apps outwith Debian and Ubuntu repos we should test and validate 'installs and runs on forky' before asserting it.  I confess to having resorted to similar confident assumptions in the past and got away with it (because the VM system running CI tests on a matrix of runners on Debians, Ubuntus, Mint and Pop! died). I'd likely merge this and hope for the best, but only two things are certain out here...

snoack:
You are *certainly* right that there is a good chance that this assumption will not hold for 100% of the packages targeted here, but...
1. The packages I'm targeting with this change are already marked compatible with `trixie` and `sid`. If they fail on `forky` I'm pretty sure their current metadata is already wrong and they also fail on `trixie` or `sid`.
2. It's probably better to risk introducing a couple false positives, rather than preventing ~26 packages from being installed on  `forky` for no reason.
3.  If we move forward with #1798, this decision will be automatic in the future, as these packages would have specified `trixie+` (or an earlier min codename) to start with.

PR #1801, opened 2026-03-17
feat: minimum codenames

### LINKED ISSUE #1798: Feature Request: Allow minimum release syntax in CODENAMES_SUPPORTED
### New Feature

Allow specifying minimum supported distributions in `CODENAMES_SUPPORTED`:

```
CODENAMES_SUPPORTED=">=noble >=trixie"
```

instead of requiring every supported codename to be explicitly listed.

### Background

Currently, `deb-get` package definitions require explicitly enumerating every supported Ubuntu and Debian codename, for example:

```
CODENAMES_SUPPORTED="jammy noble trixie forky sid"
```

This becomes increasingly cumbersome as well as fragile over time as new releases are added. Many upstream `.deb` packages support all current and future releases starting from a certain version, making it unnecessary to maintain a growing list of codenames.

Moreover, once new Debian/Ubuntu releases come out, `deb-get` needs to catch up with it's hard-coded list of codenames. For example, many (if not most) packages cannot be installed on Debian Forky (current *Testing*) because the package definitions haven't updated yet.

silentJET85:
closes #1798
This will allow you to append a `+` to a codename in the `CODENAMES_SUPPORTED` variable, to specify it as the minimum supported codename. This way most packages won't need to be updated every six months when a new Ubuntu is released.

For example instead of:
`CODENAMES_SUPPORTED="trixie forky sid noble questing resolute"` 
you could instead do:
`CODENAMES_SUPPORTED="trixie+ noble+"`

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -1,7 +1,4 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2076 # We often use =~ to match strings rather than a regex.
-# shellcheck disable=SC2155 #  The exit codes aren't being used in these instances anyway.
-
 LC_ALL=C
 PACKAGE_INSTALLATION_TRIES=0
 PACKAGE_INSTALLATION_COUNT=0
@@ -318,7 +315,7 @@ function upgrade_apt() {
 }
 
 function upgrade_only_dg() {
-    mapfile -t INSTALLED_APT_PPA < <(grep -P -o "\S+(?=\s\d+\s(apt\s*$|ppa\s*)$)" "${ETC_DIR}/installed")
+    mapfile -t INSTALLED_APT_PPA < <(grep -E "apt\s*$|ppa\s*$" "${ETC_DIR}/installed"); INSTALLED_APT_PPA=(${INSTALLED_APT_PPA[@]%% *})
     printf '%s\0' "${INSTALLED_APT_PPA[@]}" | xargs -0 ${ELEVATE} apt-get -q -o Dpkg::Progress-Fancy="1" -y install --only-upgrade
 }
 
@@ -484,17 +481,13 @@ function validate_deb() {
     if [ "${APP_SRC}" == "00-builtin" ]; then
         deb_"${APP}" 2>/dev/null
     else
-        # shellcheck source=/dev/null
         . "${ETC_DIR}/${APP_SRC}.d/${APP}" 2>/dev/null
     fi
 
     if [[ "${CODENAMES_SUPPORTED}" =~ "+" ]]; then
-        ALL_CODENAMES="$(distro-info --supported)"
-        if [[ "${UPSTREAM_ID}" == "debian" ]]; then
-            ALL_CODENAMES="$(distro-info --all | grep -B1 -m1 "${ALL_CODENAMES}" | head -n1) ${ALL_CODENAMES}"
-        fi
+        ALL_CODENAMES="$(distro-info --all)"
         ALL_CODENAMES="$(echo ${ALL_CODENAMES/experimental})"
-        MIN_CODENAME="$(grep -P -o "[a-z]+(?=\+)" <<< "${CODENAMES_SUPPORTED}" | grep -E -m1 "$(tr ' ' '|' <<< "${ALL_CODENAMES}")")"
+        MIN_CODENAME="$(grep -P -o "[a-z]+(?=\+)" <<< "${CODENAMES_SUPPORTED}" | grep -E -m1 "${ALL_CODENAMES// /|}")"
         if [ -n "${MIN_CODENAME}" ]; then
             CODENAMES_SUPPORTED="$(grep -E -o "${MIN_CODENAME}.*" <<< "${ALL_CODENAMES}" || echo "${CODENAMES_SUPPORTED//+}")"
         else
@@ -754,7 +747,7 @@ function refresh_supported_cache_lists() {
         fi
     fi
 
-    trap 'trap - EXIT; ${ELEVATE} rm -f "$lockfile"' EXIT
+    trap "trap - EXIT; ${ELEVATE} rm -f \"$lockfile\"" EXIT
     ${ELEVATE} touch "$lockfile"
     ${ELEVATE} rm -f "${CACHE_DIR}/supported.list" "${CACHE_DIR}/supported_apps.list"
     if [ -z ${QUIET} ]; then
@@ -767,7 +760,6 @@ function refresh_supported_cache_lists() {
     cut -d" " -f 1 "${CACHE_DIR}/supported.list" | sort -u | ${ELEVATE} tee "${CACHE_DIR}/supported_apps.list" >/dev/null
 }
 
-
 function update_repos() {
     local REPO_URL=""
     # preserve current behaviour for now but allow modification via env
@@ -775,7 +767,6 @@ function update_repos() {
     local WGET_VERBOSITY="--quiet"
 
     if [[ " $* " == *' --quiet'* ]]  ; then
-        # shellcheck disable=SC2034
          CURL_OPTS="--disable --show-error --silent"
     fi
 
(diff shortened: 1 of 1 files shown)


deb-get
```diff
@@ -476,14 +471,35 @@ function validate_deb() {
...
     # Source the variables
     if [ "${APP_SRC}" == "00-builtin" ]; then
         deb_"${APP}" 2>/dev/null
     else
         . "${ETC_DIR}/${APP_SRC}.d/${APP}" 2>/dev/null
     fi
+
+    if [[ "${CODENAMES_SUPPORTED}" =~ "+" ]]; then
+        local DEBIAN_CODENAMES="bullseye bookworm trixie forky sid"
+        local UBUNTU_CODENAMES="jammy noble questing resolute"
```
snoack: You can get a complete list of all codenames for the current distribution in release order from `distro-info --all`. You would need to make `deb-get`'s own *deb* package depdend on the `distro-info` package though. But since `deb-get` is distributed as a *deb* package anyway and `distro-info` is widely available on all supported distributions/releases, I think adding that dependency is well worth it in order to not have to maintain this hard-coded list. What do you think?

silentJET85: It sounds like a good idea, assuming that the command still returns the Debian and Ubuntu codenames even in the various forks of Debian/Ubuntu.

EDIT: After looking into it a bit, it appears that that command gets it's data from a csv file in /usr/share/distro-info/. If we parse that file directly, we wouldn't even need to add the dependency.
The files seem to be installed by the package `distro-info-data`. I've checked a few of my VMs and it seems to already be installed by default in all of them so far.

snoack: > It sounds like a good idea, assuming that the command still returns the Debian and Ubuntu codenames even in the various forks of Debian/Ubuntu.

I don't think you need to bother with foreign distribution's codenames. Let's say for example we have `CODENAMES_SUPPORTED="noble+ bookworm+"` in our package definition and we are running on Debian Trixie. Now, `distro-info --all` returns `... bookworm trixie ... sid`, so we substitute  `bookworm+` with `bookworm trixie ... sid` (i.e. the output of `distro-info --all` with everything before `bookworm` removed). `noble+` (or any unknown codename to be expanded) can just be substituted with an empty string (or left as-is) since it will never match the current system's codename anyway, right?

> EDIT: After looking into it a bit, it appears that that command gets it's data from a csv file in /usr/share/distro-info/. If we parse that file directly, we wouldn't even need to add the dependency.

`/usr/share/distro-info/` is provided by the `distro-info-data` package. If you read from these files directly you still might want to add a dependency to that package instead.

silentJET85: My worry was that maybe some distros might not list any of the upstream codenames at all. But it seems like that's not the case. It looks like most distros actually include both of the Debian and Ubuntu csv lists. 
And if they have their own, such as in Devuan, it's in addition to those.

silentJET85:
I think adding `distro-info-data` as a dependency is a good idea. 

It would also allow us to altogether remove the need to manually add support for new codenames as they release. Though I'll save that for a different pull request.

snoack:
The new syntax should probably be documented in `EXTREPO.md`.

deb-get
```diff
@@ -478,6 +478,27 @@ function validate_deb() {
...
+        DEBIAN_CODENAMES="${DEBIAN_CODENAMES#*$'\n'}"; DEBIAN_CODENAMES="${DEBIAN_CODENAMES%$'\n'}"; DEBIAN_CODENAMES="${DEBIAN_CODENAMES//$'\n'/ }"
+        local UBUNTU_CODENAMES="$(cut -d ',' -f 3 "/usr/share/distro-info/ubuntu.csv")"
+        UBUNTU_CODENAMES="${UBUNTU_CODENAMES#*$'\n'}"; UBUNTU_CODENAMES="${UBUNTU_CODENAMES%$'\n'}"; UBUNTU_CODENAMES="${UBUNTU_CODENAMES//$'\n'/ }"
+        local MIN_CODENAMES="$(grep -P -o "[a-z]+(?=\+)" <<< "${CODENAMES_SUPPORTED}")"
+        MIN_CODENAMES="${MIN_CODENAMES%$'\n'}"; MIN_CODENAMES="${MIN_CODENAMES//$'\n'/|}"
+        if [ -n "${MIN_CODENAMES}" ]; then
+            local OTHER_CODENAMES="$(grep -E -o "[a-z]+(\s|$)" <<< "${CODENAMES_SUPPORTED}")"
+            OTHER_CODENAMES="${OTHER_CODENAMES//$'\n'}"
+            CODENAMES_SUPPORTED="$(grep -E -o "(${MIN_CODENAMES})[^,]*" <<< "${DEBIAN_CODENAMES},${UBUNTU_CODENAMES},")"
+            CODENAMES_SUPPORTED="${CODENAMES_SUPPORTED/$'\n'/ }"; CODENAMES_SUPPORTED="${CODENAMES_SUPPORTED% }"
```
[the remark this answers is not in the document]
silentJET85: @cubic-dev-ai This was already addressed by the `if [ -n "${MIN_CODENAMES}" ]` block. 
If `${MIN_CODENAMES}` is empty, then `${CODENAMES_SUPPORTED}` is never reassigned.

philclifford:
Just a thought: I wonder if making the dependency on distro-info and using e.g. `echo $(${UPSTREAM_ID}-distro-info --supported)` is an even better option : that way we get dynamically listed supported releases and not just new ones and don't have to bother with parsing the csv directly.

silentJET85:
Yeah, I think that could work. 

And as a bonus, having the distro-info command available could come in handy to use in some packages. Such as when we need to fall back to a deb file or repo meant for a previous release. 
That way we don't have to hardcode which codenames to fallback to, or just blindly decrementing UPSTREAM_RELEASE and hoping for a match, as I've been doing.

silentJET85:
I've found one problem with the `--supported` flag. It doesn't consider releases in Debian's LTS support period to be "supported." In other words, Bullseye is missing.
```
jet@devuan:~$ distro-info --supported
bookworm
trixie
forky
sid
experimental
```

EDIT: I came up with a workaround. If we run `distro-info` a second time, but with the `--all` flag, we can grep for the previous output and grab the line that occurs before it. It should be safe to assume that the line preceding the list of "supported" releases will be the one currently in LTS support.

silentJET85:
@cubic-dev-ai Can you re-review this?

silentJET85:
This needs one final tweak. Using `distro-info --supported` to get our list of distros, results in `CODENAMES_SUPPORTED` being blank if a package has an outdated release set as it's minimum.

EDIT: I have changed it to use `distro-info --all` instead. I think this is the simplest solution. 

The only side effect would be that some packages could potentially continue working even when running on an EOL distro. 
In packages where `CODENAMES_SUPPORTED` is unset, that would happen anyway. So it doesn't seem like a big deal.

deb-get
```diff
@@ -484,9 +481,20 @@ function validate_deb() {
...
     fi
+
+    if [[ "${CODENAMES_SUPPORTED}" =~ "+" ]]; then
+        ALL_CODENAMES="$(distro-info --all)"
+        ALL_CODENAMES="$(echo ${ALL_CODENAMES/experimental})"
+        MIN_CODENAME="$(grep -P -o "[a-z]+(?=\+)" <<< "${CODENAMES_SUPPORTED}" | grep -E -m1 "${ALL_CODENAMES// /|}")"
+        if [ -n "${MIN_CODENAME}" ]; then
+            CODENAMES_SUPPORTED="$(grep -E -o "${MIN_CODENAME}.*" <<< "${ALL_CODENAMES}" || echo "${CODENAMES_SUPPORTED//+}")"
+        else
+            CODENAMES_SUPPORTED="$(grep -P -o -m1 "[a-z]+(?=\s|$)" <<< "${CODENAMES_SUPPORTED}")"
```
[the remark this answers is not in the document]
silentJET85: @cubic-dev-ai This is intentional. In the event where `$MIN_CODENAME` is unset and there are no other codenames without a `+`, then that means we _want_ all codenames to be supported. 

The scenario where this would likely happen is if a package definition only specifies a minimum codename for Debian, but specifies no codenames at all for Ubuntu. This would cover the common occurrence where the oldest Debian release lacks some dependency needed by an app, but Ubuntu doesn't. No need to specify a minimum for both, if it isn't actually needed.

PR #1802, opened 2026-03-17
feat: allow "all" archs in packages

### LINKED ISSUE #1797: Feature Request: Support ARCHS_SUPPORTED="all" in package definitions
### New Feature

Allow `ARCHS_SUPPORTED="all"` in package definitions to indicate that a package supports all architectures, instead of requiring every architecture to be listed explicitly.

### Background

Currently, deb-get package definitions require explicitly listing all supported architectures in the ARCHS_SUPPORTED variable, e.g. `ARCHS_SUPPORTED="amd64 arm64 armhf"`.

However, many .deb packages are architecture-independent or are published for every architecture supported by Debian/Ubuntu. In these cases the Debian package metadata often uses `Architecture: all`.

For such packages, repository definitions could be simplified if deb-get allowed `ARCHS_SUPPORTED="all"`.

silentJET85:
closes #1797

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -484,7 +484,7 @@ function validate_deb() {
     else
         . "${ETC_DIR}/${APP_SRC}.d/${APP}" 2>/dev/null
     fi
-    if [[ " ${ARCHS_SUPPORTED} " =~ " ${HOST_ARCH} " ]] && { [ -z "${CODENAMES_SUPPORTED}" ] || [[ " ${CODENAMES_SUPPORTED} " =~ " ${UPSTREAM_CODENAME} " ]]; } && { [ "${METHOD}" != ppa ] || [ "${UPSTREAM_ID}" == ubuntu ]; }; then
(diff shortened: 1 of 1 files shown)


philclifford:
It seemed so innocuous but seems to contain or trigger a sneaky logic problem :-(

philclifford:
compare 

```  
./deb-get prettylist 01-main |grep debian |head

```

between a610e25acf0672c2a6e9a468dcd11c5ffd3f27b3 and 586c8668bb36a2140201f7ab26346669e10602d8

I suspect it may break more than the listing but CI didn't test updating or (re)installing a few apps from each source ...

silentJET85:
This puzzled me for a while. 
It turns out that neither `$HOST_ARCH` nor "all" were matching. I'm surprised that installing apps somehow continued to work. After trying lots of things, I was starting to think that pipes and parenthesis don't work in combination with quoted strings. I ended up making a commit that avoided it by doing a separate [[ ]] test just for "all". That works, but is a bit ugly. 

Then I finally realized that my original method was fine. The problem was simply some spaces around the pipe. Doh! :facepalm:

PR #1823, opened 2026-03-26
feat: add support for Deb822 (.sources files)

### LINKED ISSUE #1816: Feature Request: Support .sources files
### New Feature

Since Debian Trixie, `apt` supports `.sources` files replacing old `.list` files, and shows a message to run `apt modernize-sources` to convert your old `.list` files. `deb-get` only supports `.list` files which isn't ideal.

silentJET85:
closes #1816
EDIT: (I reworked this. See latest posts down below.)

This adds optional support for the newer Deb822 `.sources` files. Existing support for `.list` files is not affected.

~~These new variables are used in a package to create a `.sources` file:~~
~~* `APT_SOURCES_URL` This contains the URL to the repository. This variable being set is also what tells deb-get to use the new format.~~
~~* `APT_SUITES` This contains the suite(s). Equivalent to the "word" after the URL in the old format.~~
~~* `APT_COMPONENTS` The component(s). In the old format, this would usually be the remainder of the line after the "suite." If the suite ends with a `/`, this is unused.~~

These existing variables are also used:
* `APT_LIST_NAME` This variable is still used to create the name of the file, but with the `.sources` extension.
~~* `APT_REPO_OPTIONS` This still contains the options, but you need to specify them in the new format. 
 For example `"Architectures: amd64"` instead of `"arch=amd64"`~~
* `ASC_KEY_URL`, `GPG_KEY_URL` and `GPG_KEY_ID` work exactly the same as normal.
 
 ~~For some examples of how these would be used, here are some modified versions of existing packages that I used for testing:~~

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -1323,12 +1323,12 @@ function add_apt_repo() {
         fi
         APT_SOURCES_LINES+=("Signed-By: /usr/share/keyrings/${APT_LIST_NAME}-archive-keyring.gpg")
         if [[ -e "/etc/apt/sources.list.d/${APT_LIST_NAME}.list" ]]; then
-            fancy_message info "Removing /etc/apt/sources.list.d/${APT_LIST_NAME}.list"
-            ${ELEVATE} rm "/etc/apt/sources.list.d/${APT_LIST_NAME}.list"
+            fancy_message info "Renaming /etc/apt/sources.list.d/${APT_LIST_NAME}.list to ${APT_LIST_NAME}.list.bak"
+            ${ELEVATE} mv "/etc/apt/sources.list.d/${APT_LIST_NAME}.list" "/etc/apt/sources.list.d/${APT_LIST_NAME}.list.bak"
         fi
         printf '%s\n' "${APT_SOURCES_LINES[@]}" | ${ELEVATE} tee "/etc/apt/sources.list.d/${APT_LIST_NAME}.sources" > /dev/null
 
-    else
+    elif ! [[ -e "/etc/apt/sources.list.d/${APT_LIST_NAME}.sources" ]]; then
         # single line .list format
         local APT_LIST_LINE="deb [signed-by=/usr/share/keyrings/${APT_LIST_NAME}-archive-keyring.gpg"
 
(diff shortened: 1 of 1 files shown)


silentJET85:
Actually, now I'm reconsidering if this is the best approach. Parsing the existing `APT_REPO_URL` and `APT_REPO_OPTIONS` variables for the same information seems like it would be pretty easy. And then the package definition files wouldn't need to change at all. 
~~I think we could make it so that new installs of apps would use the new format by default, while already installed apps would keep using `.list` files.~~
The safest option is probably to keep using `.list` files by default, and only switch to the new format if an environment variable is set. I'll make another commit that uses this method. I've got it mostly completed.

silentJET85:
I've updated it to check an environment variable called `DEBGET_SOURCES`. If set to `sources`, the .sources format will be used for all packages that use the `apt` method. Otherwise, a .list file will be used. In the future, we could easily swap this to make the new format default.
This seems to work fine for `apt` packages. But I haven't tested PPAs yet. I'll leave this as a draft until I do.
EDIT: Yes, PPAs are working.

philclifford:
Just a thought.  You check for an existing .list when the variable turn on .sources behaviour and remove it.  This is fine (I'd rename it to something that would be ignored but others seem to just leave it in place (presumably the .sources takes precedence or both are parsed and only generate errors on conflict).  

```
$ ls -l /etc/apt/sources.list.d/google-chrome.{list,sources} /etc/apt/sources.list.d/vscode.{list,sources}
ls: cannot access '/etc/apt/sources.list.d/vscode.list': No such file or directory
-rw-r--r-- 1 root root 190 Mar 22 23:23 /etc/apt/sources.list.d/google-chrome.list
-rw-r--r-- 1 root root 313 Apr 21 09:30 /etc/apt/sources.list.d/google-chrome.sources
-rw-r--r-- 1 root root 278 Apr 16 19:57 /etc/apt/sources.list.d/vscode.sources

```

However there's nothing in the default .list behaviour to check for an existing .sources file.  I think there probably should be because

- upstream might handle the migration via packaging and we don't want to stamp on that.
- at some point users might get to run the handy `sudo apt modernize-sources` 
- Users may have configured the same repository using other tooling for another package in the same repository
- The DE822 file might have important additional control information 

We absolutely should not remove or rename the DEB822 file if present.

The question then is what do we do if a .sources files exists ? Error or warn? Skip the .list and just use the .sources ?

silentJET85:
I think the overall cleanest solution would be to use the existing .sources file if it already exists. 


> * Users may have configured the same repository using other tooling for another package in the same repository

This is tricky. If we avoid creating our own .sources file, since one already exists, our app could end up orphaned if the user uninstalls the first app and it has a postrm script to remove its repository. 

On the other hand, if we add our own .sources file, in addition to the existing one, apt will always complain about the repository being configured multiple times.

The third option is to add our own .sources file and delete the original. (Presumably it would use a different filename.) But then we have the opposite problem. If the deb-get app is uninstalled, the non-deb-get app could end up orphaned.

Maybe a solution could be: Don't add our own .sources if one already exists, but whenever we run `deb-get update` we check that each app with the `apt` method still has a valid repo, and if not, re-add it.

But now that I think about it, these scenarios don't actually seem specific to .sources files. I think it's already possible this could happen.

philclifford:
This is too tricky to get into the release that is about to be made.  
It is not forgotten nor is it unappreciated - just complicated.

PR #1835, opened 2026-04-03
fix(fsearch): separate trixie from unstable repo mapping

### LINKED ISSUE #1808: Bug: fsearch won't install in Debian 13 or Unstable
### Expected Behavior

fsearch installs correctly.

### Actual Behavior

There are a few issues. 
* The package definition file currently has Trixie trying to use Unstable's repo, which is now too new for it.
*  They still don't have a repo for Trixie. The Bookworm repo seems to work fine in Trixie though. So this isn't a big deal.
* The Unstable repo cannot be installed in Debian Unstable. It's because apt now rejects repos that are signed with a v3 signature. OBS didn't add v4 support until late last year, and fsearch hasn't had a release since then. I suspect this problem will fix itself whenever they put out a new release. One possible workaround would be to grab the [deb file](https://download.opensuse.org/repositories/home:/cboxdoerfer/Debian_Unstable/amd64/fsearch_0.2.3-1+4.17_amd64.deb) manually, but it's probably better to just wait.

mvanhorn:
## Summary

Separates Trixie/Debian 13 from sid in the fsearch package definition's case statement, so Trixie uses the Bookworm (12) OBS repo instead of Unstable.

## Why this matters

The current mapping (`sid|13|trixie -> Unstable`) causes fsearch installation failures on Debian 13 because the Unstable OBS repo uses v3 GPG signatures that newer apt versions reject. The Bookworm repo works correctly on Trixie (confirmed by @silentJET85 in #1808).

## Changes

- Split `sid|13|trixie) DEBIANVER=Unstable` into two cases:
  - `sid) DEBIANVER=Unstable` (preserves existing behavior for sid)
  - `13|trixie) DEBIANVER=12` (uses working Bookworm repo)

## Testing

CI package tests will verify the definition is valid.

Fixes #1808

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/01-main/packages/fsearch
+++ b/01-main/packages/fsearch
@@ -1,4 +1,5 @@
 DEFVER=2
+CODENAMES_SUPPORTED="bullseye bookworm jammy noble"
 # shellcheck disable=SC2168
 local THE_KEY
 # shellcheck disable=SC2168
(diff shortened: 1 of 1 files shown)


philclifford (changes requested):
We should not kludge this but rather remove support for UPSTREAM_RELEASE where UPSTREAM_CODENAME has a native release upstream as suggested in #1808.

Then the case statement is no longer needed and DEBIANVER can just become UPSTREAM_RELEASE

mvanhorn:
Reworked per your feedback. Commit 52cba92 removes the UPSTREAM_CODENAME case statement entirely and adds `CODENAMES_SUPPORTED="bullseye bookworm jammy noble"` so trixie, sid, and newer Ubuntu releases are dropped - fsearch is in the native repos for those now. DEBIANVER is just UPSTREAM_RELEASE directly.

Based on the #1808 discussion where @silentJET85 confirmed trixie and testing both resolve to the native Debian package, and you agreed to de-support those releases.

philclifford (approved):

PR #1836, opened 2026-04-07
Add Cable package definition

### LINKED ISSUE #1677: App Request: Cable
### Is this a new request?

### App Name

Cable

### Download URL

https://github.com/magillos/Cable/releases

### Website URL

https://github.com/magillos/Cable

### App Summary

PyQT GUI application to dynamically modify [PipeWire](https://pipewire.org/) and [WirePlumber](https://pipewire.pages.freedesktop.org/wireplumber/) settings at runtime, such as quantum, sample rate, audio profiles, latency offset, services restart and more. It features side-by-side, MIDI matrix and graph style connections managers (uses Python Jack Client so will not list Pipewire items), pw-top wrapper, simple ALSA mixer and jack_delay GUI.

nrocha02:
Closes #1677

### DIFF
--- a/01-main/packages/cable
+++ b/01-main/packages/cable
@@ -1,4 +1,5 @@
 DEFVER=1
+CODENAMES_SUPPORTED="trixie forky sid noble questing resolute"
 get_github_releases "magillos/Cable" "latest"
 if [ "${ACTION}" != prettylist ]; then
     URL=$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
--- a/01-main/packages/cable
+++ b/01-main/packages/cable
@@ -2,7 +2,7 @@ DEFVER=1
 get_github_releases "magillos/Cable" "latest"
 if [ "${ACTION}" != prettylist ]; then
     URL=$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
-    VERSION_PUBLISHED=$(cut -d '-' -f 2 <<< "${URL/.deb/}")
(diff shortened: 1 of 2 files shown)


01-main/packages/cable
```diff
@@ -0,0 +1,9 @@
+DEFVER=1
```
philclifford, line 1:
```suggestion
DEFVER=1
CODENAMES_SUPPORTED="trixie forky sid noble questing resolute"
```

philclifford: Dependencies do not resolve on jammy / bookworm.

01-main/packages/cable
```diff
@@ -0,0 +1,9 @@
...
+get_github_releases "magillos/Cable" "latest"
+if [ "${ACTION}" != prettylist ]; then
+    URL=$(grep -m 1 "browser_download_url.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
+    VERSION_PUBLISHED=$(cut -d '-' -f 2 <<< "${URL/.deb/}")
```
philclifford:
```suggestion
    VERSION_PUBLISHED=$(cut -d '/' -f 8 <<< "${URL//v/}")
```
This pattern reliably gets the tagged version (and yes the removal of the 'v' is not needed here, but is harmless and matches ~150 existing definitions which might be candidates for a little re-factoring) rather than cutting the filename, which is quite prone to upstreams making naming tweaks and breaking definitions.

philclifford (changes requested):
The packaging dependencies do not resolve on older LTS supported releases so support must be restricted to working releases.
Also it is better to adopt the pattern of getting the version tag from the asset path rather than cutting the filename.

philclifford (approved):

PR #1843, opened 2026-04-12
fix: "show" not showing Download and Published

### LINKED ISSUE #1842: Bug: "deb-get show" missing info for non-installed packages
### Expected Behavior

All lines populated when doing `deb-get show`.

### Actual Behavior

>Unless I'm overlooking something, there doesn't seem to be any need to have the cache file for non-installed apps.
_Originally posted by @silentJET85 in https://github.com/wimpysworld/deb-get/issues/1419#issuecomment-3018034286_

I finally found one side effect of #1419. 
If a package using `get_github_releases`, `get_gitlab_releases` or `get_website` is not installed, the `Download:` and `Published:` lines will be blank when you do a `deb-get show` for that package.
```
jet@debian13:~$ deb-get show gb-studio
  [*] WARNING! Cached file /var/cache/deb-get/gb-studio.json_extract is empty or missing.
GB Studio
  Package:      gb-studio
  Repository:   01-main
  Updater:      deb-get
  Installed:    No
  Published:
  Architecture: amd64 arm64
  Download:
  Website:      https://www.gbstudio.dev/
  Summary:      A quick and easy to use drag and drop retro game creator.
```

silentJET85:
closes #1842

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -180,7 +180,7 @@ function get_github_releases() {
     #   curl -I https://api.github.com/users/flexiondotorg
 
     # Do not process github releases while generating a pretty list or upgrading
-    if [[ ' install fix-installed ' =~ " ${ACTION} " ]] || { [[ ' update ' =~ " ${ACTION} " ]] && package_is_installed "${APP}"; }; then
+    if [[ ' install fix-installed show ' =~ " ${ACTION} " ]] || { [[ ' update ' =~ " ${ACTION} " ]] && package_is_installed "${APP}"; }; then
         if [ ! -e "${CACHE_FILE}" ] || [ -n "$(find "${CACHE_FILE}" -mmin +"${DEBGET_CACHE_RTN:-60}")" ]; then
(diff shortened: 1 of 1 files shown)


silentJET85:
I found another problem. `deb-get reinstall` also will fail if the `CACHE_FILE` doesn't already exist. I went ahead and added the fix to this PR as well.

philclifford:
Oops - fixed by #1848.  Sorry I missed your PR.

philclifford:
> I found another problem. `deb-get reinstall` also will fail if the `CACHE_FILE` doesn't already exist. I went ahead and added the fix to this PR as well.

That's not in #1848. I wonder if it is really needed because `reinstall` should only happen if a package is already installed, in which case the CACHE_FILE should generally exist already.  Personally I think the "solution" to users trying to reinstall a package they don't have already installed would be to check in reinstall and either silently call install instead or output a grumpy message telling the user to do that themselves.

I might be wrong ...

PR #1848, opened 2026-04-17
fix: populate release cache on 'show' for non-installed packages

### LINKED ISSUE #1842: Bug: "deb-get show" missing info for non-installed packages

mvanhorn:
Closes #1842

## Problem

`deb-get show <pkg>` prints empty `Published:` and `Download:` lines when the package is not installed and uses `get_github_releases`, `get_gitlab_releases`, or `get_website`:

```
jet@debian13:~$ deb-get show gb-studio
  [*] WARNING! Cached file /var/cache/deb-get/gb-studio.json_extract is empty or missing.
GB Studio
  Package:      gb-studio
  Repository:   01-main
  Updater:      deb-get
  Installed:    No
  Published:
  Architecture: amd64 arm64
  Download:
  Website:      https://www.gbstudio.dev/
  Summary:      ...
```

Per-package scripts like `01-main/packages/gb-studio` derive `URL` and `VERSION_PUBLISHED` by reading `${CACHE_FILE}`, but since #1419 the cache is only populated for `install`, `fix-installed`, and `update` (the latter only when the package is already installed). `show` was not in that allowlist, so for non-installed packages the cache file never existed, the grep in the package definition returned nothing, and the two lines came out blank.

## Fix

Add `show` to the action allowlist in `get_github_releases`, `get_gitlab_releases`, and `get_website`. The existing "no cache for apps we never install" guarantee (no fetching during `list`, `prettylist`, `csvlist`, `search`, etc.) is preserved.

## Testing

- `bash -n deb-get` passes
- The three helpers are the only places where cache population is gated by `${ACTION}`; the same pattern is applied consistently across all three

## Scope notes

- No change for `list`, `prettylist`, `csvlist`, `search`, `remove`, `purge` — cache is still not fetched during those actions
- No change for packages that use `direct` sources — they do not depend on the cache file
- Per-hour cache reuse (`DEBGET_CACHE_RTN`) still applies, so repeated `show` calls won't hammer GitHub / GitLab APIs

### DIFF
--- a/deb-get
+++ b/deb-get
@@ -180,7 +180,7 @@ function get_github_releases() {
     #   curl -I https://api.github.com/users/flexiondotorg
 
     # Do not process github releases while generating a pretty list or upgrading
-    if [[ ' install fix-installed ' =~ " ${ACTION} " ]] || { [[ ' update ' =~ " ${ACTION} " ]] && package_is_installed "${APP}"; }; then
+    if [[ ' install fix-installed show ' =~ " ${ACTION} " ]] || { [[ ' update ' =~ " ${ACTION} " ]] && package_is_installed "${APP}"; }; then
         if [ ! -e "${CACHE_FILE}" ] || [ -n "$(find "${CACHE_FILE}" -mmin +"${DEBGET_CACHE_RTN:-60}")" ]; then
             if [ -z ${QUIET} ]; then
                 fancy_message info "Updating ${CACHE_FILE}"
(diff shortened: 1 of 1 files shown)


mvanhorn:
Thanks for the merge, @philclifford. Caching on show makes the UX a lot snappier.

PR #1849, opened 2026-04-18
feat(nexis): add Nexis system optimizer

### LINKED ISSUE #1845: App Request: Nexis
### Is this a new request?

- [x] I have checked that this app does not already have a pending request, has not previously been declined and does not already exist in deb-get.

### App Name

Nexis

### Download URL

https://github.com/s4solutionsllc/Nexis/releases

### Website URL

https://s4solutionsllc.github.io/Nexis/

### App Summary

Cross-platform system optimizer for Linux & macOS. Successor to Stacer (9k+ stars) rebuilt on Qt 6/C++17 with GPU monitoring, Docker management and disk health. GPL-3.0.

Nexis began as a fork of [Stacer](https://github.com/oguzhaninan/Stacer), a popular Linux system optimizer created by oguzhaninan. After the original project went inactive in 2020, development continued here -- porting to Qt 6 and C++17, adding native macOS support, GPU monitoring, a hardware info panel, kiosk mode, and fixing 38+ bugs inherited from the upstream codebase. As the feature-set diverged, the project was rebranded as Nexis to reflect that it had become something new. Stacer laid the foundation; Nexis is where it goes from here.

mvanhorn:
Added package manifest for Nexis, a cross-platform system optimizer that succeeds Stacer (9k+ stars), rebuilt on Qt 6/C++17.

GitHub Releases provide `.deb` packages for amd64 and arm64.

Closes #1845

### DIFF
--- a/01-main/packages/nexis
+++ b/01-main/packages/nexis
@@ -1,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64 arm64"
+get_github_releases "s4solutionsllc/Nexis" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL=$(grep "browser_download_url.*${HOST_ARCH}.*\.deb\"" "${CACHE_FILE}" | head -1 | cut -d '"' -f 4)
+    VERSION_PUBLISHED=$(cut -d '/' -f 8 <<< "${URL//v/}")
+fi
+PRETTY_NAME="Nexis"
+WEBSITE="https://s4solutionsllc.github.io/Nexis/"
+SUMMARY="Cross-platform system optimizer for Linux & macOS. Successor to Stacer, rebuilt on Qt 6/C++17."
--- a/01-main/packages/nexis
+++ b/01-main/packages/nexis
@@ -1,5 +1,6 @@
 DEFVER=1
 ARCHS_SUPPORTED="amd64 arm64"
+CODENAMES_SUPPORTED="resolute stonking trixie sid"
 get_github_releases "s4solutionsllc/Nexis" "latest"
 if [ "${ACTION}" != "prettylist" ]; then
     URL=$(grep -m 1 "browser_download_url.*${HOST_ARCH}.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
--- a/01-main/packages/nexis
+++ b/01-main/packages/nexis
@@ -1,9 +1,13 @@
 DEFVER=1
 ARCHS_SUPPORTED="amd64 arm64"
-CODENAMES_SUPPORTED="resolute stonking trixie sid"
-get_github_releases "s4solutionsllc/Nexis" "latest"
+CODENAMES_SUPPORTED="noble resolute stonking trixie sid"
+get_github_releases "s4solutionsllc/Nexis"
 if [ "${ACTION}" != "prettylist" ]; then
-    URL=$(grep -m 1 "browser_download_url.*${HOST_ARCH}.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
+    NEXIS_UBUNTU_RELEASE="ubuntu2604"
+    if [ "${UPSTREAM_CODENAME}" = "noble" ]; then
+        NEXIS_UBUNTU_RELEASE="ubuntu2404"
+    fi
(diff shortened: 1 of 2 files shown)


mvanhorn:
Added the nexis manifest entry in e893d9c. Inserted alphabetically between net.downloadhelper.coapp and nextcloud-desktop, so package discovery and release workflows pick it up.

01-main/packages/nexis
```diff
@@ -0,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64 arm64"
+get_github_releases "s4solutionsllc/Nexis" "latest"
+if [ "${ACTION}" != "prettylist" ]; then
+    URL=$(grep "browser_download_url.*${HOST_ARCH}.*\.deb\"" "${CACHE_FILE}" | head -1 | cut -d '"' -f 4)
```
philclifford:
```suggestion
    URL=$(grep -m 1 "browser_download_url.*${HOST_ARCH}.*\.deb\"" "${CACHE_FILE}" | cut -d '"' -f 4)
```

01-main/packages/nexis
```diff
@@ -0,0 +1,10 @@
+DEFVER=1
+ARCHS_SUPPORTED="amd64 arm64"
```
philclifford, line 2:
```suggestion
ARCHS_SUPPORTED="amd64 arm64"
CODENAMES_SUPPORTED="resolute stonking trixie sid"
```
deb is built for resolute and lower versions are directed to PPA or APPImage 
see https://github.com/s4solutionsllc/Nexis/releases/tag/v2.7.2
and  
https://github.com/s4solutionsllc/Nexis/blob/native/RELEASE.md#currently-supported-targets

mvanhorn:
Fixed in b8f1b14 - noble dropped from CODENAMES_SUPPORTED since Nexis only ships ubuntu2604 artifacts. ShellCheck and the codename smoke check pass.

philclifford:
> Fixed in [b8f1b14](https://github.com/wimpysworld/deb-get/commit/b8f1b14d60e02edf8068046ad6054b2ea2263795) - noble dropped from CODENAMES_SUPPORTED since Nexis only ships ubuntu2604 artifacts. ShellCheck and the codename smoke check pass.

Sadly removing noble breaks the CI, which runs on `ubuntu-latest` which currently github points to 24.04 (noble) and so the runner finds itself unsupported.

If we cannot contrive a fix to run the CI on "resolute if noble not in supported list else latest" (so it will continue to move forward as github move where latest points) or persuade github to move the latest on then CI will not pass until latest -> resolute

philclifford:
If you add back noble and plucky and change the URL match to get the latest 24.04, 25.04 or 26.04 released deb as appropriate (look at the definition for `lact` : it also need to remove the dot from UPSTREAM_RELEASE and combine UPSTREAM_RELEASE_NODOT and HOST_ARCH and UPSTREAM_ID to get the asset name.   The definition already gets all releases not just latest so jammy, noble and resolute should find the latest supported release for them. Debian codenames will need handling to point to the appropriate ubuntu package and Mint users might also need special handling.

This approach should allow the CI to pass.

mvanhorn:
Done, following the `lact` pattern you pointed at.

`CODENAMES_SUPPORTED` gets noble and plucky back, and the suffix is now derived (`UPSTREAM_RELEASE_NODOT` combined with `UPSTREAM_ID`) instead of hardcoding 2604 with a noble special case. Since `get_github_releases` pulls all releases rather than latest, each codename lands on the newest release that still ships a matching deb:

```
noble    -> ubuntu2404  (newest is v2.5.0)
plucky   -> ubuntu2504  (newest is v2.6.1)
resolute -> ubuntu2604  (v2.8.3)
```

stonking, trixie and sid are pinned to ubuntu2604, since upstream publishes no 2610 or debian asset.

On Mint: I don't think it needs special handling. deb-get already falls back to `ID_LIKE` and sets `UPSTREAM_ID=ubuntu`, so a Mint box on a noble base resolves through the same path and picks up ubuntu2404. Happy to add something if you've seen it behave otherwise.

The manifest entry the bot flagged is already there at `01-main/manifest:246`.

mvanhorn:
Thanks for shepherding the Nexis addition through, @philclifford. Nice to have it in the catalog.

PR #1898, opened 2026-06-03
fix: deprecate open-work

philclifford:
They no longer build deb artifacts, only AppImage since they moved from Tauri to Electron.  We need to deprecate.


philclifford:
Keep an eye on https://github.com/different-ai/openwork/issues/2084
as we could revert/re-support if the [PR](https://github.com/different-ai/openwork/pull/2087) is merged

Issue #1890, opened 2026-05-29
bug: CI reports success on failed package installation


philclifford:
See  https://github.com/wimpysworld/deb-get/actions/runs/26349341030/job/77564741758?pr=1886#logs

the package definition as submitted  in #1886  fails but the CI test "passes", as does the result reporting.

philclifford:
Also on (hopefuly) fixing the package definition the CI automatic re-run detects another package as "changed" (gram)

philclifford:
<img width="1567" height="795" alt="Image" src="https://github.com/user-attachments/assets/6b4b558c-e1af-48da-85f5-36a3b520e91a" />

silentJET85:
If an install fails due to missing info, deb-get is returning an exit code of `0`. 
As a test I edited a local override of `bat` to make `URL` and `VERSION_PUBLISHED` empty:

```
jet@devuan:~$ deb-get install bat 
  [+] Including local package bat
  [!] Please consider contributing back new entries, an issue (or raise a PR) directly at https://github.com/wimpysworld/deb-get/pulls
  [!] ERROR! Missing required information of github package bat:
URL=
VERSION_PUBLISHED=
jet@devuan:~$ echo $?
0
```

silentJET85:
As for the CI detecting other packages, I think this is the relevant part of [tests-01-main.yml](https://github.com/wimpysworld/deb-get/blob/main/.github/workflows/tests-01-main.yml)

```
if [ -n "${{ github.event.pull_request.base.sha }}" ]; then
            # Pull request: detect changed package files
            CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep "^01-main/packages/" | sed "s|^01-main/packages/||" || true)
```
`github.event.pull_request.base.sha` refers to the commit of main at the time the PR was originally created. So any packages added since then would be seen as changed.

philclifford:
Yes, I think you're right and we should be getting the diff to HEAD of main not the PR base.

philclifford:
Fixed in d5b3d64

PR #1900, opened 2026-06-03
fix: return exit code 1 when validation fails

silentJET85:
~~fixes~~ partially addresses #1890

### DIFF
--- a/.github/workflows/tests-01-main.yml
+++ b/.github/workflows/tests-01-main.yml
@@ -198,7 +198,7 @@ jobs:
               fi
             fi
           else
-            STATUS="skip"
+            STATUS="fail"
             ERROR_MSG="Package not in supported list"
           fi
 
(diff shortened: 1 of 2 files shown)


philclifford:
Thanks - it helps but there is more to it than that.

Check 
https://github.com/wimpysworld/deb-get/actions/runs/26889998850/job/79312994322?pr=1899#step:5:1361

The package is not listed in the manifest so is not installed (not in supported list) and the CI appears to be saying fail but the result is reported as a pass

silentJET85:
Added a commit that makes the CI fail instead of skip when the manifest entry is missing. It seems to work correctly in my fork. It looks like deprecated packages are already filtered out by this point, so I _think_ there shouldn't be any side effects from this.

https://github.com/silentJET85/deb-get/actions/runs/27364943617/job/80861875877?pr=29

PR #1964, opened 2026-08-12
Check for jq installation before running wget command to get github releases

shemgp:
Figured out why in a VPS `deb-get` was not installing:  `jq` was not installed! So here's a quick fix that will check for 'jq' installation and display the proper error message so people would know what to do next.

<a href="https://cubic.dev/pr/wimpysworld/deb-get/pull/1964?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #1527, opened 2025-09-25
Bug: Can not install Docker Desktop


MH0386:
### Expected Behavior

Install Docker Desktop

### Actual Behavior

```
❯ deb-get show docker-desktop
  [!] ERROR! Missing required information of website package docker-desktop:
URL=https://desktop.docker.com/linux/main/amd64/206054/docker-desktop-amd64.deb
VERSION_PUBLISHED=
```

### Steps to reproduce

`❯ deb-get install docker-desktop`

### /etc/os-release

```ini
PRETTY_NAME="Ubuntu 25.04"
NAME="Ubuntu"
VERSION_ID="25.04"
VERSION="25.04 (Plucky Puffin)"
VERSION_CODENAME=plucky
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=plucky
LOGO=ubuntu-logo
```

### /etc/lsb-release

```ini
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=25.04
DISTRIB_CODENAME=plucky
DISTRIB_DESCRIPTION="Ubuntu 25.04"
```

### deb-get version

```bash
0.4.5
```

### deb-get update

```bash
[+] Updating /etc/deb-get/01-main
####################################################################################################################################################################################### 100.0%
  [+] Updating cache of supported apps in the background
Hit:1 https://downloads.flox.dev/by-env/stable/deb stable/ InRelease
Hit:2 https://brave-browser-apt-release.s3.brave.com stable InRelease
Hit:3 https://download.docker.com/linux/ubuntu plucky InRelease
Hit:4 http://eg.archive.ubuntu.com/ubuntu plucky InRelease
Hit:5 https://packages.microsoft.com/repos/code stable InRelease
Hit:6 http://eg.archive.ubuntu.com/ubuntu plucky-updates InRelease
Hit:7 https://cli.github.com/packages stable InRelease
Hit:8 http://eg.archive.ubuntu.com/ubuntu plucky-backports InRelease
Hit:9 https://mirror.mwt.me/shiftkey-desktop/deb any InRelease
[...]

PR #1975, opened 2026-08-17
fix(docker-desktop): use appcast.json instead of scraping release notes

jarrodcolburn:
# fix(docker-desktop): use appcast.json instead of scraping release notes

`docker-desktop` is currently broken on `main` — both `URL` and `VERSION_PUBLISHED`
extraction fail, so the package errors out and is skipped:

```
$ deb-get show docker-desktop
  [!] ERROR! Missing required information of website package docker-desktop:
URL=https://desktop.docker.com/linux/main/amd64/224084/docker-desktop-amd64.deb
VERSION_PUBLISHED=
```

## Two separate failures

**1. `VERSION_PUBLISHED` is empty — this is a regression.**

The current definition uses `awk 'match($0, /re/, a)'`. Three-argument `match()` is a
gawk extension; Debian and Ubuntu ship **mawk** as the default `awk`, which errors:

```
awk: line 1: syntax error at or near ,
```

This is the exact symptom of #1527, which was fixed in 204423ea by switching to
`grep -oP`. That fix was then reverted by 753bface ("chore: standardise coding",
#1546), which restored the broken awk form while removing quotes for style. The bug
has been live on `main` since 2026-03-21.

**2. The scraped `URL` is malformed and 403s.**

`grep -m 1 -o 'https:[^>]*\.deb'` runs greedily across markup on the release-notes
page and drops a path separator:

```
scraped:   https://desktop.docker.com/linux/main/amd64236836/docker-desktop-amd64.deb  → 403
corrected: https://desktop.docker.com/linux/main/amd64/236836/docker-desktop-amd64.deb → 200
```

## Why switch source rather than re-fix the regex

`https://docs.docker.com/desktop/release-notes/` is a rendered docs page, and parsing
it has needed repair repeatedly — 76a32b6c, 61e4c63c, ae633be3, 61a655c4, b60d6ac7
(2022), ace55e44 (2024), cec53f46 (2024), 3ab7cd93 (2025), 204423ea (2025), and now
again. Each fix holds until Docker restyles the page.

`https://desktop.docker.com/linux/main/amd64/appcast.json` is the feed Docker
Desktop's own updater consumes. It is public, unauthenticated, stable in shape, and
gives the version, build number and artifact URL as data rather than markup:

```json
{
  "Items": [
    { "Visibility": "100", "AppVersion": "4.86.0", "BuildNumber": "236216",
      "Artifacts": [ { "URL": "…/236216/docker-desktop-amd64.deb", "Type": "deb" } ] },
    { "Visibility": "99",  "AppVersion": "4.87.0", "BuildNumber": "236836",
      "Artifacts": [ { "URL": "…/236836/docker-desktop-amd64.deb", "Type": "deb" } ] }
  ]
}
```

Two details this exposes that a scrape cannot:

- **Staged rollout.** Entries are oldest-first and carry a `Visibility` percentage.
  The newest build is neither the first entry nor the one at `Visibility: 100`, so
  the definition selects on highest `BuildNumber`. Picking either of the other two
  would currently offer 4.86.0 as an "update" over an installed 4.87.0.
- **`VERSION_PUBLISHED` now matches the package.** `AppVersion-BuildNumber` is
  byte-identical to the `.deb`'s own `Version` field (`4.87.0-236836`). The old bare
  `4.87.0` compared as *older* than an installed `4.87.0-236836` under
  `dpkg --compare-versions`, so a Docker rebuild of the same version was silently
  missed. `deb-get show` also now displays Installed and Published in one format.

Parsing is done with `jq`. That's already a hard requirement: the README lists it in
the install line (`sudo apt install curl lsb-release wget jq`), `get_github_releases`
pipes through it, and b153804 (#1964) made it explicit four days ago with a
`command -v jq` check that errors and exits.

A line-oriented `awk`/`grep` version was tried first and rejected: it depends
on the feed staying pretty-printed, and against a minified copy of the same document it
returned `1.0-1.0` — a silently *wrong* version rather than an empty one. Parsing the
structure removes that whole failure class, which is the point of the change.

## Testing

Ubuntu 24.04 (noble), amd64, deb-get 0.4.7:

```
$ deb-get show docker-desktop
Docker Desktop
  Package:      docker-desktop
  Published:    4.87.0-236836
  Download:     https://desktop.docker.com/linux/main/amd64/236836/docker-desktop-amd64.deb
```

Download URL returns 200 and its SHA256 matches the `Checksum` field in the appcast.

- Selection is order-independent — reversing `Items` yields the same result.
- Identical output against a minified copy of the feed.
- `BuildNumber` is compared with `tonumber`, so build 100 sorts above build 99.
- Single-item feed resolves correctly.
- Empty `Items`, truncated JSON, and non-JSON responses all yield empty values, which
  trip the existing "Missing required information" guard rather than producing a bad
  URL.
- `deb-get prettylist` unaffected (exit 0, row renders).
- `shellcheck` clean.

<a href="https://cubic.dev/pr/wimpysworld/deb-get/pull/1975?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/01-main/packages/docker-desktop
+++ b/01-main/packages/docker-desktop
@@ -1,9 +1,13 @@
 DEFVER=1
 ARCHS_SUPPORTED="amd64"
-get_website "https://docs.docker.com/desktop/release-notes/"
+# The appcast has no "latest" entry; every live release is listed, oldest-first, each
+# with a staged-rollout "Visibility" percentage. Neither document order nor
+# Visibility==100 identifies the newest build, so select on highest BuildNumber.
+# AppVersion-BuildNumber matches the .deb's own Version field.
+get_website "https://desktop.docker.com/linux/main/amd64/appcast.json"
 if [ "${ACTION}" != "prettylist" ]; then
-    URL=$(grep -m 1 -o 'https:[^>]*\.deb' "${CACHE_FILE}" )
-    VERSION_PUBLISHED=$(awk 'match($0, /href=#[0-9][^>]*>([^<]*)</, a) { print a[1]; exit }' "${CACHE_FILE}")
+    URL=$(jq -r '.Items | max_by(.BuildNumber|tonumber) | .Artifacts[]? | select(.Type=="deb") | .URL' "${CACHE_FILE}" 2>/dev/null)
+    VERSION_PUBLISHED=$(jq -r '.Items | max_by(.BuildNumber|tonumber) | select(.) | "\(.AppVersion)-\(.BuildNumber)"' "${CACHE_FILE}" 2>/dev/null)
 fi
 PRETTY_NAME="Docker Desktop"
 WEBSITE="https://www.docker.com/products/docker-desktop/"


jarrodcolburn:
Kept this PR to the bug itself. If it's accepted, there are two follow-ups I'd be
happy to pick up separately — flagging them here so the scoping is visible rather
than arriving unannounced.

**1. The `docker-ce-cli` dependency / Docker's apt repository.**

Docker Desktop's `Depends` is:

```
curl, qemu-system-x86 (>= 5.2.0), docker-ce-cli, libseccomp2, libcap-ng0, pass,
desktop-file-utils, libgtk-3-0, libx11-xcb1, uidmap
```

Nine of those resolve from the distro archive. Only **`docker-ce-cli`** comes from
Docker's own repository — it ships `/usr/bin/docker`, which Docker Desktop's
`postinst` symlinks to `/usr/local/bin/docker`. That one dependency is why
[Docker's install docs](https://docs.docker.com/desktop/setup/install/linux/) open
with "Set up Docker's `apt` repository" on both the
[Ubuntu](https://docs.docker.com/desktop/setup/install/linux/ubuntu/) and
[Debian](https://docs.docker.com/desktop/setup/install/linux/debian/) pages.

Without it, `deb-get install docker-desktop` aborts on an unmet dependency. It fails
safe — nothing half-installs — but the apt error doesn't hint that deb-get already
ships a `docker-ce` package configuring exactly that repository. An `install`-guarded
`fancy_message warn` pointing at `deb-get install docker-ce` covers it in a few lines.
It would want to work for both IDs, the way `docker-ce` already does with
`${UPSTREAM_ID}` in its `APT_REPO_URL`.

I had this written and tested but pulled it out to keep this PR a pure bugfix — happy
to open it once this lands, or fold it in here if you'd prefer it in one go.

**2. `CODENAMES_SUPPORTED`.**

The definition currently declares none, so it offers itself on every Debian/Ubuntu
release. Docker supports a narrower set — Ubuntu 24.04/26.04 LTS, Debian 12 — so
pinning the codenames would stop deb-get advertising it where Docker doesn't support
it. Worth confirming against the docs at the time rather than hardcoding from today's
page.

**On the other distributions:** the same docs cover RHEL, Fedora and Arch, and the
appcast this PR switches to carries `rhel`, `rpm` and `zst` artifacts alongside the
`deb` one. Those are outside deb-get's remit — it's Debian/Ubuntu and derivatives —
so there's nothing to do here beyond noting that the same feed would serve an
equivalent tool on those platforms, no scraping required.

philclifford (approved):
Excellently explained and executed.  Many thanks.

philclifford:
Re 1. `docker-ce-cli`above

I look forward to a further PR to implement your mitigation of the possible failure if they do not have the docker repository enabled.

For experienced users advising the prior installation of `docker-ce` to get the apt repository enabled would work.   

Re 2. pinning to supported LTS codenames

This is nuanced.  Users who are on non-LTS paths would probably expect to be able to install and have the "appropriate" prior LTS-based updates available to them. If we include mapping logic in a `case` statement but only list supported LTS codenames in the definition with interims commented out and an explanatory note it would be trivial for users to copy and amend the definition to override the strict pinning to LTS support.  Or we could be strict.  Or we could not limit by codename as now - iirc there is only one .deb and it works fine on noble+ if their apt repo is in place....

PR #2009, opened 2026-09-09
fix: show error and fatal messages raised by package definitions

### LINKED ISSUE #1997: bug: `warn` writes to stdout and survives; `error` and `fatal` write to `>&2` and are swallowed.
**P2 — the wasted download is real. The suggested fix does not achieve it.**

You are right that the warning does not stop the install; the ~450MB still transfers and
apt still fails. I had claimed otherwise in this PR's description and have corrected it.

But `return` from the definition does not skip the install. `return` in a sourced file
ends the *sourced file*, not the sourcing function, so `validate_deb` continues at the
line after the `.`. Tested both placements:

- `return 1` **before** `PRETTY_NAME`/`SUMMARY`/`WEBSITE` — sourcing stops, those stay
  empty, and `validate_deb` fails its own required-field check. That does abort, but by
  way of `ERROR! Missing required information of package docker-desktop`, which reads
  like a broken definition rather than a missing dependency.
- `return 1` **after** those are set — no effect at all; `deb-get show` renders normally
  and the install proceeds.

`fancy_message fatal` is the intended mechanism and does abort cleanly before the
download, but its message is invisible: definitions are sourced at deb-get:489 with
stderr discarded.

```bash
. "${ETC_DIR}/${APP_SRC}.d/${APP}" 2>/dev/null
```

`warn` writes to stdout and survives; `error` and `fatal` write to `>&2` and are
swallowed. Confirmed by emitting both from a definition — only the `warn` appeared. So
`fatal` from a definition is a silent `exit 1`.

That leaves no way for a definition to abort *and* explain itself, which is a core-level
constraint rather than something this package should work around. It also means
`winegui`'s `fancy_message error` on line 24 currently prints nothing. I have left the
behaviour as a warning and kept the fix to one line of core out of scope for this PR —
happy to send it separately if @wimpysworld and @philclifford want it.

_Originally posted by @jarrodcolburn in https://github.com/wimpysworld/deb-get/issues/1984#issuecomment-5463644579_

alorentiar:
Definitions are sourced with stderr discarded, so fancy_message error and
fatal messages raised while loading them never made it to the user.
fatal in particular was just a silent exit 1, which is why the
docker-desktop definition ended up using a warning instead (see #1984).

This flags the sourcing window in validate_deb and has error/fatal fall
back to stdout while it is set, so a definition can abort and explain
itself. Everything outside that window still goes to stderr as before.

Kept the existing stderr suppression in place - dropping it entirely
surfaces unrelated noise from a few definitions (e.g. tofu's unquoted
ARCHS_SUPPORTED line), which is a separate problem.

Tested locally on Ubuntu 24.04 with throwaway definitions in 99-local.d:
a load-time error now prints its reason, and a fatal aborts with the
message visible. prettylist over all definitions is unchanged (0 bytes
stderr), shellcheck reports nothing new.

Closes #1997


alorentiar:
Good catch - prettylist and csvlist output is parsed downstream, so writing the message to stdout there was wrong.

The sourcing window in validate_deb now keeps the real stderr on fd 3, and error/fatal messages go there instead. They still reach the terminal, just not through stdout. Tested with a throwaway definition that raises at load time: prettylist/csvlist stdout stays clean (no error text in the markdown/CSV) while the message shows up on stderr. Normal runs are unchanged - prettylist over all definitions still exits 0 with 0 bytes of stderr.

alorentiar:
Valid point - thank you. validate_deb no longer touches the shell's fd 3 directly. The redirection is now scoped to the sourcing command itself (`. file 3>&2 2>/dev/null`, same for the 00-builtin branch), so fd 3 points at the real stderr only while the definition is being sourced and is restored to whatever the caller had before, untouched.

Verified both sides again with a throwaway definition that raises at load time: error/fatal messages still land on stderr (visible, and prettylist/csvlist stdout stays clean), and an fd 3 opened before validation is still usable afterwards.

PR #2019, opened 2026-09-12
chore(i3): align i3 repo support with supported ubuntu releases

philclifford:


philclifford:
CI fails on noble because of signature but
```
$ gpg --no-default-keyring --keyring /usr/share/keyrings/sur5r-keyring.gpg --verify InRelease.txt
gpg: Signature made Mon 19 Jan 2026 09:35:34 AM GMT
gpg:                using RSA key BFD90F4DAAEFA72B67BBAF48E3CA1A89941C42E6
gpg: Good signature from "debian.sur5r.net Archive Automatic Signing Key 2015 (sur5r) <debian@sur5r.net>" [unknown]
```

At this point I'm minded to just deprecate i3

